PR #110, opened 2026-08-25
Randomise office spawn heading and report yaw relative to takeoff

S3333F:
## Description

Office telemetry handed every policy an absolute world-frame yaw, and every episode spawned facing the same direction - a compass a real drone does not have. A survey-based policy can use that to localise against a baked map without ever looking at the camera. The spawn heading is now drawn per episode from its own seeded stream, and the reported yaw is relative to the takeoff heading, which is what a real IMU gives you after it zeroes on the pad.

Fixes # (issue)

### Related Tasks

- Task ID: 61

### Type of Change

### What does this PR change?

- Spawn: the chaser starts at a random heading drawn from `seed ^ OFFICE_HEADING_SEED_OFFSET`. The offset gets its own stream on purpose: the env writes the accepted spawn back into the shared task object, so two envs built from the same seed can consume a different number of placement draws - a heading riding that stream would differ between them.
- The yaw controller target is initialised to the spawn heading before the first step; without it the controller snaps the drone back to world zero.
- Telemetry: the returned yaw is relative to the takeoff heading. Body-frame velocity and acceleration keep using the true physical yaw - only the reported angle is relative.
- Docs and the submission template now state the attitude row as relative-to-takeoff, no world compass.
- Tests: spawn-heading randomisation and relative-yaw reporting; a body-accel frame test that fails if acceleration were converted with the relative yaw (fixture seed spawns at ~262 degrees, so the wrong frame puts forward thrust sideways); an aim-at-target preamble for the two detector tests that assumed the camera faces the target at spawn.

### How was this tested?

- Commands run: `python3 -m pytest tests/ -q` - 973 passed, 76 skipped in 5:06 (skips are the opt-in heavy tests and a capnp schema absent in this environment, both unrelated to this change). Determinism check: two envs built from the same seed produce identical episodes.
- Family proofs re-run with the change: solo navigation 100/100, scripted-pilot chase 38/40 (the script's ceiling).
- Current top office submission re-evaluated on 25 seeds: 0.827 on `main` vs 0.812 with this change. Its baked-map clearance veto still works; the drop is the heading no longer being free.

### Tools & Dependencies

- Tools updated: None

Miners: state indices 2-3 (the sin/cos yaw pair) now encode yaw relative to takeoff, and the spawn heading varies per episode. A policy that treated yaw as world-absolute, or relied on the fixed spawn heading to align a baked survey, will see its score move.

> [!WARNING]
> Same-seed episodes differ from `main`. This has to ship with a version bump at an epoch boundary so every validator evaluates on the same worlds; the bump is not in this PR.

### Risk & Rollback

If the relative-yaw frame is wrong somewhere, policies get a rotated attitude channel and scores drop across the board. Two tests guard the frames: the acceleration test catches body acceleration projected with the relative instead of the physical yaw, and the spawn-heading test checks reported yaw starts at zero on the pad. Revert the commit to undo; no state or data involved.

### Documentation

- [x] `docs/` updated

### Additional Information

Pre-existing quirk left untouched: the env writes the accepted spawn back into the shared task object, so a second env built from the same task starts its placement from the accepted spawn. It is a harmless fixed point today; the heading stream is isolated from it either way.

### DIFF
--- a/docs/families/office_interceptor.md
+++ b/docs/families/office_interceptor.md
@@ -66,7 +66,7 @@ The 15 telemetry values mirror a Tello state packet: only what the physical SDK 
 
 | Slice | Channel | Contents |
 |-------|---------|----------|
-| 0–3 | `attitude_pitch_roll_sincos_yaw` | pitch, roll (rad, 1° steps), sin(yaw), cos(yaw) |
+| 0–3 | `attitude_pitch_roll_sincos_yaw` | pitch, roll (rad, 1° steps), sin(yaw), cos(yaw) — yaw is **relative to the takeoff heading**, which is seeded per episode: like a real IMU, there is no world compass |
 | 4–6 | `body_velocity_xyz` | forward / right / down velocity (m/s, 0.1 steps = the SDK's dm/s) |
 | 7–9 | `body_acceleration_xyz` | forward / right / down specific force (m/s², with sensor bias) — **gravity included**: a hovering drone reads ≈ −9.8 on the down axis, like the real IMU |
 | 10–14 | `altitude_tof_height_baro_age_valid` | downward ToF (m; never returns the target — a target underneath reads as the floor), fused height (m), barometer with drift (m), packet age (s), valid flag |
--- a/swarm/__init__.py
+++ b/swarm/__init__.py
@@ -18,7 +18,7 @@
 import sys
 from pathlib import Path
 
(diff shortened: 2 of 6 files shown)


miguelik2 (approved):
LGTM

robojules (approved):
LGTM

PR #116, opened 2026-08-31
Deal the office a different room, airframe and target every episode

S3333F:
## Description

The office family shipped a fixed world: one floorplan, one set of control constants, one target size, and a spawn distribution concentrated enough to be worth memorising. A policy could therefore win by fitting our simulator rather than by flying, and the current champion does exactly that — it carries a copy of our office as an occupancy grid, reimplements our stick-to-velocity mapping to invert it, hardcodes the target's dimensions to turn box size into range, and lists room coordinates ordered by spawn probability.

Five properties are now dealt per episode from their own seeded streams, so none of them is a constant worth memorising, and the catch rule is rewritten around the geometry of a real interception. The changes were chosen against the physical Tello this family exists to transfer to: an airframe's response varies with battery and wear, a site is never the size of the drawing, and no interceptor approaches from directly overhead.

Fixes # (issue)

### Related Tasks

- Task ID: egBWsvHD

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Catch rule | Interception geometry replaces a centre-distance sphere | Miners: overhead approaches stop scoring, level passes start scoring reliably |
| Airframe | Speed, dead zone, slew rate and motor lag dealt +/-18% per episode | Miners: a fitted inverse model stops transferring |
| Detector | Target's apparent size dealt +/-10% per episode | Miners: box size no longer inverts to a known range |
| Map | Every room axis stretched 2-5% per episode | Miners: a memorised metric floorplan no longer aligns |
| Placement | Both drones drawn uniformly over reachable free space | Miners: no spatial prior over spawn location |
| Scoring | Par time derived from the airframe the episode dealt | Miners: a slow draw is not judged on a fast drone's clock |

- **Catch.** `OFFICE_KILL_RADIUS_M` is replaced by `OFFICE_CATCH_RADIUS_M` (0.20 m horizontal), `OFFICE_CATCH_LEVEL_M` (0.08 m vertical) and `OFFICE_CATCH_HOLD_STEPS` (3). Success is a real hull contact, or holding inside that envelope for three consecutive control steps. The hulls (r = 0.088 m) touch at 0.176 m, so the old 0.15 m sphere could never fire on a level pass — every level intercept depended on a contact being sampled inside one 20 ms step. From above, the flat hull (0.042 m tall) let the centre distance fall inside 0.15 m untouched, which paid for hovering over the target and descending onto it.
- **Airframe.** `office_airframe_profile(map_seed)` is a pure seed-to-airframe function on its own stream. The env fits itself from it and `_calculate_office_target_time` derives par from the same profile, so the clock a seed is judged against matches the drone it was given. The flee allowance stays on the nominal speed, since the target's flee speed is not randomised.
- **Detector.** The episode's silhouette is drawn on its own stream and `_detector_capture` builds box width and height from it rather than from module constants.
- **Map.** `office_scale(seed)` returns a per-axis factor applied as `meshScale` to every office body. The flyable bounds, the ceiling, the window plug, the target's waypoint sampling and the escape check all read the episode's room. The placement navigation grid is built against the room actually standing and cached on that room's exact scale, because the clearance radius does not stretch with the room and a component joined at one size can be sealed at another.
- **Placement.** `_clear_spawn` draws every candidate over free space instead of nudging the task's suggested point, and no longer reads the task at all — the env writes its accepted placement back into the shared task object, so anchoring on it made a second env built from the same task consume a different number of draws. It returns only points that passed clearance, the sealed-room check and the 4-14 m separation band, and raises otherwise.

### How was this tested?

- Commands run: `python3 -m pytest tests/ -q`
- Result: 986 passed, 76 skipped in 11:06. The skips are the opt-in heavy tests and a capnp schema absent in this environment, both unrelated.

The current champion (UID 142) was re-flown against the change on a render box, on the same seeds before and after. It is the policy this PR is aimed at, so it is the honest measure of whether the handles were removed without breaking the task.

<details>
<summary>UID 142 re-flown, same 10 seeds, before and after</summary>

```
                caught   mean score   median catch   median path
before (main)   10/10       0.9958         6.2 s        11.7 m
after           10/10       0.9801         9.2 s        16.1 m
```

A wider run of 16 seeds on the new family: 16/16 caught, mean score 0.9876, median catch 6.5 s, no failures. Room scales spanned -4.8% to +4.8%, dealt airframes 2.48 to 3.25 m/s, target widths 16.7 to 19.7 cm.

The catch rule specifically does not touch this policy: over both runs it ends the episode level with the target (median height difference 0.041 m) and makes hull contact in 7 to 8 of 10 episodes, spending a median 0% of the flight above it. What the change costs it is precision rather than capability — it flies 4.4 m further and 3 s longer for the same catch.

</details>

Determinism was checked explicitly, since every property here is drawn: `test_office_telemetry_deterministic` builds two envs from one seed and compares the full observation stream, and `test_office_map_deterministic` builds one seed's room twice and compares every body's AABB.

### Tools & Dependencies

- Tools updated: None

Miners: an interception must now be flown alongside the target rather than dropped onto it, and five properties that were previously fixed vary per episode. Nothing in the observation or action layout changes and no resubmission is needed, but a policy that fitted the previous constants will score differently. The family guide and the starter template state the new ranges and catch semantics.

> [!NOTE]
> The overhead catch was not a theoretical hole. The current champion ships the routine that exploits it — a terminal elevation guard that holds 0.11 m above the target, with a comment stating it "lets the centre distance fall inside 0.15 m, which registers instantly and with no impulse" — gated behind an environment variable that defaults to off.

### Risk & Rollback

The randomisation bands are deliberately narrow and every draw is seeded, so the failure mode is a task that is slightly harder than intended rather than one that is non-deterministic; the two determinism tests above are the guard. Revert the commits; nothing here is stateful and no other family reads these constants.

### Documentation

- [x] `docs/` updated


miguelik2 (approved):
LGTM, announcement and deploy

robojules (approved):
LGTM

PR #117, opened 2026-08-31
Run only the miner tests when a pull request only touches miner files

S3333F:
## Description

Every pull request runs the whole suite, about seven minutes, whether it touched the simulator or a
line in the miner's starter controller. A pull request whose files are all miner-facing now runs the
tests that cover those files and nothing else, in about ten seconds. Everything else runs the suite
exactly as before.

The direction of the decision is the point: a file nobody has said anything about runs the whole
suite. Running everything costs minutes, while running the miner tests alone over a change they do
not cover reports a pass that was never checked.

Stacked on #115 and based on that branch, so the diff here is only the routing.

Fixes # (issue)

### Related Tasks

- Task ID: CQ7QCpGp

### Type of Change

- [ ] Bug fix
- [x] Performance
- [ ] Documentation

### What does this PR change?

- `ci/miner_tests.toml` lists the eight files a miner-only change may touch and names the tests
  covering each. Paths are exact and there are no patterns, so a file added tomorrow is not on the
  list and runs the whole suite until somebody says what covers it. `bounds` records where the miner
  side ends, so widening it is a deliberate edit rather than a side effect of adding a row.
- `ci/classify_changes.py` reads a `git diff --name-status -z -M` payload and prints `miner` or
  `full`. Only a plain modification to a listed file takes the short lane: an add, a delete, a
  rename or a file becoming a symlink changes what those tests were written against. Anything it
  cannot parse is `full`.
- `ci/run_miner_tests.py` runs what the manifest names, so the workflow never carries a second copy
  of the list that could drift from the first.
- `.github/workflows/tests.yml` picks the lane and runs it in one job. The classifier and the
  manifest are read from the first parent of the merge commit rather than from the branch, so a
  branch cannot widen the rules it is about to be judged by, and the lane reaches the shell through
  `env:` rather than expression interpolation.
- `.docker/docker-compose.yml` gains a `swarm_miner` service running the same entry point, so the
  lane can be reproduced locally.

One job rather than several is deliberate. Splitting the lanes across jobs needs a gate job to
report on their behalf, and a job skipped by `if:` reports as success, so the gate ends up speaking
for work that never ran. With one job the check that is required is the one that invokes pytest.

### How was this tested?

- Full suite on this branch, in the CI container as the runner's non-root user:

```
pytest -q -n4 --dist worksteal -p no:cacheprovider
1053 passed, 76 skipped, 96 warnings in 379.02s (0:06:19)
```

- The miner lane, run exactly as the workflow runs it:

```
python ci/run_miner_tests.py
55 passed, 1 skipped, 12 warnings in 9.65s
```

  The skip is `tests/test_submission_manifest.py:127`, which wants a `swarm-backend` checkout beside
  this one; it skips on `main` too.

- The lane choice is tested by pulling the workflow's own shell out of the YAML and running it
  against merge commits built for the purpose, so the test exercises what CI will execute rather
  than searching the file for a string:

| Change on the branch | Lane |
|---|---|
| a listed miner file | `miner` |
| a listed miner file and a validator file | `full` |
| a validator file, plus a manifest rewritten to allow it | `full` |
| a validator file, plus a classifier replaced with one that prints `miner` | `full` |

- `tests/test_miner_tests_manifest.py` also fails if a listed file does not exist, if a selector
  collects no tests, if a row falls outside `bounds`, or if a routing file is ever listed.

- `tests/test_starter_templates.py` runs the starter controllers under `python -I` from outside the
  checkout. Run from inside it, a starter that imports `swarm` passes a test meant to prove it
  stands on its own.

### Tools & Dependencies

- Tools updated: None.

No change for miners or validator operators. Contributors see a shorter wait on pull requests that
only touch miner files.

### Database & API

### Risk & Rollback

If the classifier is wrong in the safe direction a pull request runs the full suite, which is what
happens today. If it were wrong the other way a change would be merged on tests that do not cover
it, so every unhandled case returns `full` and the tests above pin the ones worth naming. Revert the
commits and the workflow runs the whole suite for everything again.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the manifest carries its own comments, and contributors do not
  interact with the routing directly.

### Before merging

- [ ] Repoint the required status check from `gate` to `tests` in branch protection. The job name
      changes with this PR, and a required check that no longer exists is reported as missing.

### DIFF
--- a/.docker/docker-compose.yml
+++ b/.docker/docker-compose.yml
@@ -20,3 +20,16 @@ services:
     command: >
       pytest -m "not slow and not integration"
              -n 2 --dist worksteal -ra -p no:cacheprovider
+
+  # The miner slice, for the CI lane that runs when only miner files changed.
+  # Reads the same manifest the workflow does, so the two cannot disagree.
+  swarm_miner:
+    <<: *swarm-base
+    container_name: swarm_miner
+    restart: "no"
+    working_dir: /app
+    volumes:
+      - ..:/app
+    environment:
+      SWARM_TERRAIN_CACHE_DIR: /tmp/swarm_terrain_cache
+    command: python ci/run_miner_tests.py
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -4,6 +4,8 @@ on:
   pull_request:
   push:
     branches: [main]
+  merge_group:
+    types: [checks_requested]
 
 # a new push cancels the run already going
 concurrency:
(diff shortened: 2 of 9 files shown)


robojules (approved):
LGTM

S3333F:
Closing: these commits landed on main through #118, which is based on this branch. The lane this PR added was replaced there by two jobs, one for miner/tests and one for tests, both running on every pull request.

PR #118, opened 2026-08-31
Move the miner's own code into a miner folder

S3333F:
## Description

The miner's own code was spread across `neurons/`, `RL/`, `scripts/` and `swarm/`, so
nothing in the repository said what belongs to a miner and what belongs to the validator.
It now sits under `miner/`, with its code, its tests and its guide together.

Fifteen of the sixteen moves are renames git tracked on its own. What made this worth
doing carefully is the other kind of breakage: a path written relative to the file that
holds it means something different once the file moves, and three of those failed
silently rather than turning a test red.

Stacked on #117 and based on that branch, so the diff here is only the migration.

Fixes # (issue)

### Related Tasks

- Task ID: k4gjzzgg
- Related tasks/PRs (other repos): follows #115 and #117

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

**The layout**

```
miner/
  src/     miner.py, drone_agent.py, office_drone_agent.py, RL/, scripts/
  tests/   test_miner.py, test_starter_templates.py, test_repo_root_paths.py
  docs/    miner.md
```

**What did not move, and why**

| Stayed | Reason |
|---|---|
| `swarm/submission_template/` less the two starters | `batch.py:1342` stages `main.py`, `agent_server.py`, `runtime_caps.py` and `agent.capnp` into every submission; `_shared.py:59` resolves that directory and the evaluation image copies it |
| `swarm/submission_manifest/` | imported by `swarm/cli.py` |
| `swarm/templates/README.md` | pinned by hash, see below |

**The starter README, and what it costs to move a page it links to**

`swarm/templates/README.md` ships byte-for-byte into every miner repository and links to
the guide and the training starters by absolute URL. Moving either page means correcting
those links, which changes the README's sha256.

The backend already supports that. `REQUIRED_README_HASH` admits a new submission, and
`ACCEPTED_README_HASHES` holds every template the project has enforced, so a repository
published against an older one stays eligible. Its own comment: *"A king's repo keeps the
template it was published with... admission still requires `REQUIRED_README_HASH`, so a
new submission cannot ship an outdated template."*

So the links were corrected, `REQUIRED_README_HASH` moved with them, and the two pages
that existed only to answer the old URLs are gone.

> [!IMPORTANT]
> This needs a matching change in `swarm-backend`, deployed before or with this merge:
> set `REQUIRED_README_HASH` to `9139c39669a9865c597861e09130b3cb57a6d9a293829ef0732c27d78af3c669`
> and prepend the previous value `952cda7c086215460006e627056cab287e72138fedc4129156e335cc9f5ad0da`
> to `ACCEPTED_README_HASHES` in `app/github_utils.py`. Without it the backend admits the
> old template and rejects the new one, which is the wrong way round. Existing kings stay
> eligible either way; miners re-run `swarm repo package` before their next submission.

**Three paths the move broke**

| File | Was | Became |
|---|---|---|
| `miner/src/miner.py` | `parent.parent` reached the repository root | reached `miner/` |
| 5 × `RL/cf_*/train.py` | `parents[2]` put the root on `sys.path` | put `miner/src` there |
| `RL/test_RL.py` | `parents[1]` reached the root | reached `miner/src` |

`_load_local_families` catches every exception and returns `None`, so the first showed up
as a family list that was quietly empty. The trainers are not covered by any test, so the
other six showed up as nothing at all. `miner/tests/test_repo_root_paths.py` now pins each
of the seven walks to the repository root, and fails if a new trainer is added without
being listed.

**Coverage the move removed**

`tests/test_scripts_shell.py` found its subjects by globbing `scripts/`. Moving the
miner's two scripts out of it dropped four parametrised checks with the suite still green.
It now looks in both places and asserts the miner contributes exactly its two scripts.

**Two jobs instead of a lane**

The routing added in #117 is removed here. Choosing a lane means asserting a change is
confined to one side, and the run then does not check that assertion. The manifest already
held a counter-example: `swarm/submission_manifest` was listed as miner-only while
`swarm/cli.py` imports it, so a change there would have skipped the CLI tests.

`.github/workflows/tests.yml` now runs a `miner` job over `miner/tests` and a `validator`
job over `tests`, both on every pull request. `ci/classify_changes.py`,
`ci/miner_tests.toml`, `ci/run_miner_tests.py` and the two test files guarding them are
gone, 536 lines. Lanes can come back if the suite grows slow enough to want them.

**Settings that decide what ships and what runs**

Reverting any of these left the suite green, so each is asserted against the artifact
rather than the configuration:

- `pytest.ini` gains `miner/tests` as a test root. `test_the_miner_tests_are_actually_collected`
  runs a default `pytest --collect-only` and requires a `miner/tests` node ID, so an
  `--ignore` in `addopts` cannot pass it.
- `pyproject.toml` discovers `miner*` and excludes `miner.tests*`; `MANIFEST.in` prunes
  `miner/tests` and includes the shell scripts. `tests/test_packaging.py` builds a wheel
  and looks inside it.
- `.gitignore` follows the training output to `miner/src/RL/*/out/`.
- `.dockerignore` excludes `miner` outright. The evaluation image consumes none of it, so
  the docker cache key, which walks `swarm/`, stays correct.

### How was this tested?

- Full suite, in the CI container as the runner's non-root user:

```
pytest -q -n 12 --dist worksteal -p no:cacheprovider
1073 passed, 76 skipped, 272 warnings in 376.65s (0:06:16)
```

- Collected test IDs before and after the move, diffed rather than counted, because a
  missing test does not fail anything: **1129 -> 1136**, the difference being the new
  guards alone with nothing lost.

- The two CI jobs, run the way the workflow runs them:

```
pytest -q miner/tests                        35 passed in 75.77s
pytest -q tests -n4 --dist worksteal        988 passed, 76 skipped in 351.46s
```

  Together they collect 1099, which is what a bare `pytest` collects, so nothing falls
  between the two jobs.

- The entry point and the schema read that broke:

```
python miner/src/miner.py --help      ok
from miner.src import miner           ok
_load_local_families()                dict, 6 families
```

- Each new guard was checked by reverting what it covers:

| Change made | Result |
|---|---|
| drop `miner/tests` from `testpaths` | fails |
| add `miner.src*` to the pyproject exclude | fails |
| add `exclude-package-data` for the miner scripts | fails |
| remove the `recursive-include` for the scripts | fails |
| delete `RL/` | fails the pinned-link test |

- `swarm --help` and the packaged wheel both still carry the four validator runtime files.

### Tools & Dependencies

- Tools updated: None.

Miners run the neuron from a new path and the setup scripts from a new path. The guide and
the family pages are updated, and the commands are:

```
./miner/src/scripts/install_dependencies.sh
./miner/src/scripts/setup.sh
python miner/src/miner.py --netuid 124 ...
```

Nothing already submitted breaks: the starter controllers import only `numpy`, so no
published submission refers to anything that moved, and the pinned README is unchanged.

### Database & API

### Risk & Rollback

The risk in a move is a reference left behind, and the ones that do not fail loudly are
the paths above and the documented commands. Both classes now have tests. Revert the
commits to put the files back; nothing outside the repository points at the new locations
yet.

### Documentation

- [x] `docs/` updated
- [ ] README.md updated
- [x] Not applicable (explain why): the guide moved and its links were repointed rather
  than rewritten.

### Before merging

- [ ] Merge #115 and #117 first; this is based on #117.
- [ ] Ship and deploy the `swarm-backend` hash change described above.
- [ ] Tell miners the run and setup commands moved, and that they should re-run
      `swarm repo package` to pick up the new README before their next submission.


miner/tests/test_repo_root_paths.py
```diff
@@ -0,0 +1,46 @@
...
+]
+
+
+def test_every_trainer_is_listed_here():
+    """A new trainer has to be added above, or it goes unchecked."""
+    found = set(MINER_ROOT.glob("src/RL/cf_*/train.py"))
+    assert found == {p for p in REACH_THE_REPO_ROOT if p.name == "train.py"}
+
+
+@pytest.mark.parametrize("source", REACH_THE_REPO_ROOT, ids=lambda p: p.name)
```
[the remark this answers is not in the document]
S3333F: Fixed in 1b865e1. The ids are now the path relative to the repository root, so a failure names the trainer:

```
[miner/src/RL/cf_autopilot/train.py]
[miner/src/RL/cf_interceptor/train.py]
[miner/src/RL/cf_search_and_rescue/train.py]
[miner/src/RL/cf_swarm_autopilot/train.py]
[miner/src/RL/cf_swarm_sar/train.py]
```

tests/test_scripts_shell.py
```diff
@@ -71,7 +80,20 @@ def test_scripts_are_not_world_writable():
...
+
+
+def test_the_miner_tests_are_actually_collected():
+    """A default run has to reach miner/tests. Reading the setting is not enough:
+    an ignore in addopts leaves testpaths looking right and collects nothing."""
+    result = subprocess.run(
+        [sys.executable, "-m", "pytest", "--collect-only", "-q", "-p", "no:cacheprovider"],
+        cwd=str(REPO_ROOT), capture_output=True, text=True,
+    )
+    assert result.returncode == 0, result.stdout[-2000:]
```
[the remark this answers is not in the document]
S3333F: Fixed in 1b865e1. `addopts` is `-ra -n auto --dist worksteal`, so this was starting a worker per core to do nothing but list tests, and xdist changed the output it then searches. It now passes `-n 0`, carries a `timeout=300`, and reports stderr alongside stdout on failure.

RL/README.md
```diff
@@ -0,0 +1,8 @@
+# Training starters
+
+These moved to [`miner/src/RL/`](../miner/src/RL) along with the rest of the miner's own
+code.
+
+This page stays because the starter README that ships into every miner repository links
+here, and that README is pinned by hash: changing it would stop the backend accepting
+submissions from repositories already published.
```
robojules: Should it be moved to docs/miner.md? So we can delete this file and RL folder, as it does not add much value?

miner/src/submission_template/drone_agent.py
robojules, line 1: What about moving this file and the `office_drone_agent.py` file into a `submission_template` | `submission` | `template folder?

miner/src/submission_template/office_drone_agent.py
robojules, line 1: What about moving this file and the `drone_agent.py` file into a `submission_template` | `submission` | `template folder?

tests/test_github.py
```diff
@@ -185,3 +188,56 @@ def test_claimed_repos_persists_to_disk(tmp_path, monkeypatch):
...
+)
+
+
+def _pinned_readme_targets() -> list[str]:
+    template = Path(__file__).resolve().parents[1] / "swarm" / "templates" / "README.md"
+    return sorted(set(_CANONICAL_LINK.findall(template.read_text())))
+
+
+# A page reached only through a legacy URL is easy to leave broken.
+LANDING_PAGES = ["docs/miner.md", "RL/README.md"]
```
robojules: Will change if the RL/README.md is updated with latest request.

tests/test_packaging.py
```diff
@@ -73,3 +77,65 @@ def test_nothing_excludes_the_data_files_again(relative_path, setuptools_config)
             assert not translate_pattern(pattern).match(within), (
                 f"{relative_path} is excluded again by '{pattern}' under '{scope}'"
             )
+
```
robojules, line 48: Why do we need to add here tests for the miner?

tests/test_packaging.py
```diff
@@ -73,3 +77,65 @@ def test_nothing_excludes_the_data_files_again(relative_path, setuptools_config)
...
+    xdist a module fixture is built once per worker that touches it."""
+    required = [
+        *REQUIRED_DATA_FILES, *MINER_DATA_FILES,
+        "miner/src/miner.py", "miner/src/drone_agent.py",
+    ]
+    missing = [p for p in required if p not in wheel_contents]
+    assert missing == [], f"the wheel is missing {missing}"
+
+    shipped_tests = sorted(n for n in wheel_contents if n.startswith("miner/tests"))
+    assert shipped_tests == [], f"the wheel carries miner tests: {shipped_tests}"
```
robojules: Move miner tests to hte miner folder

tests/test_scripts_shell.py
```diff
@@ -1,21 +1,30 @@
...
 
 import subprocess
+import sys
 from pathlib import Path
 
 import pytest
 
 REPO_ROOT = Path(__file__).resolve().parent.parent
 SCRIPTS_DIR = REPO_ROOT / "scripts"
+MINER_SCRIPTS_DIR = REPO_ROOT / "miner" / "src" / "scripts"
```
robojules: Move miner tests to the miner folder

tests/test_scripts_shell.py
```diff
@@ -71,7 +80,25 @@ def test_scripts_are_not_world_writable():
...
+        [sys.executable, "-m", "pytest", "--collect-only", "-q",
+         "-p", "no:cacheprovider", "-n", "0"],
+        cwd=str(REPO_ROOT), capture_output=True, text=True, timeout=300,
+    )
+    assert result.returncode == 0, (
+        f"stdout:\n{result.stdout[-2000:]}\nstderr:\n{result.stderr[-2000:]}"
+    )
+    assert "miner/tests/test_miner.py::" in result.stdout, (
+        "a default run does not collect the miner's tests"
+    )
```
robojules, line 95: Move miner tests to the miner folder

robojules (changes requested):
PTAL (Please take a look).

RL/README.md
S3333F: Done in c55577e — both RL/README.md and the RL folder are gone, and so is the docs/miner.md landing page.

You were right that they added little. I had kept them because the starter README links to both by absolute URL and is pinned by hash. Checking the backend showed that is already handled: REQUIRED_README_HASH admits a new submission, and ACCEPTED_README_HASHES holds every template the project has enforced, so a repo published against an older one stays eligible.

So the README now points at miner/docs/miner.md and miner/src/RL, and REQUIRED_README_HASH moved with it. This needs the matching backend change — new hash in, old hash prepended to ACCEPTED_README_HASHES — which is written up in the PR description.

miner/src/submission_template/drone_agent.py
S3333F: Done in c55577e — both are now in miner/src/submission_template/.

tests/test_github.py
S3333F: Both landing pages are gone in c55577e, so this test is now just the miner guide's own links plus the pinned README's targets.

tests/test_packaging.py
S3333F: Moved in c55577e. The miner packaging assertions are in miner/tests/test_packaging.py; tests/test_packaging.py keeps only the validator data files. The fixtures they share, including the wheel build, moved to the root conftest so the wheel is built once for both.

tests/test_packaging.py
S3333F: Moved to miner/tests/test_packaging.py in c55577e.

tests/test_scripts_shell.py
S3333F: Moved to miner/tests/test_scripts_shell.py in c55577e, which covers the miner's two scripts on their own. tests/test_scripts_shell.py is back to walking scripts/ only.

tests/test_scripts_shell.py
S3333F: One test stayed in tests/ deliberately: test_the_miner_tests_are_actually_collected checks that miner/tests is collected at all, so if it lived in miner/tests it would disappear along with everything it is meant to protect. Everything else moved in c55577e.

conftest.py
robojules, line 1: Shouldn't this file be moved to the tests folder? Thx 🙏

robojules (changes requested):
LGTM — just one minor comment to address.

conftest.py
S3333F: It has to stay at the root. pytest only applies a conftest.py to tests at or below its own directory, so moving it into tests/ would leave miner/tests without it. I tried it to be sure:

```
miner/tests: 30 passed, 5 errors
ERROR miner/tests/test_packaging.py::test_manifest_selects_the_miner_scripts
ERROR miner/tests/test_packaging.py::test_the_miner_package_is_discovered
ERROR miner/tests/test_packaging.py::test_the_miner_tests_do_not_ship
ERROR miner/tests/test_packaging.py::test_the_wheel_carries_the_miner_side
```

It also carries the repo-root sys.path setup and the --run-full / --run-e2e marker skipping, which miner/tests would lose the same way.

The file itself is not new here, it is already on main. This PR only appends the three packaging fixtures to it, and they sit there rather than in either tests folder because both roots now ask questions about the same distribution: keeping them shared means the wheel is built once instead of once per folder.

robojules (approved):
LGTM, please review that everything works so everything stills working after merging 🙏

conftest.py
robojules: Good cach btw, LGTM :rocket:

PR #119, opened 2026-09-01
Download the champion of a specific challenge family

S3333F:
## Description

`swarm champion` asks the backend for the champion without naming a challenge family, and the
backend answers that with the highest-scoring champion across all six. Today that is the interceptor
at 0.9992, so a miner working on autopilot or search-and-rescue downloads a model for a family they
are not competing in, with nothing in the output saying so.

`--family-id` names the family, and the backend already accepts it on both `/champion` and
`/models/{uid}/download` — no backend change is needed. The auto-download inside `swarm benchmark`
had the same gap and now follows the `--family-id` it was given.

Fixes # (issue)

### Related Tasks

- Task ID: Shf0HzTs

### Type of Change

- [x] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

- `swarm champion` takes `--family-id`, constrained to `CHALLENGE_FAMILY_IDS`, so `--help` lists the
  six valid ids and an unknown one is rejected by argparse before any request is sent.
- Both requests the command makes carry the family. `/models/{uid}/download` resolves a bare UID to
  the newest submission for that UID across families, so scoping only the first call leaves the two
  free to disagree once a miner submits under one UID to more than one family.
- The default output name becomes `champion_{family_id}_UID_{uid}.zip` when a family is given.
  Pulling three families' champions into one directory previously wrote all three to
  `champion_UID_{uid}.zip`. Without `--family-id` the old name is unchanged.
- `swarm benchmark` with no `--model` downloads the champion of the family it is about to benchmark.
  Previously it fetched the cross-family best, so `--family-id cf_search_and_rescue` benchmarked the
  interceptor champion on SAR maps.
- `tests/test_cli.py` covers the command for the first time: family reaches both requests, the
  no-family path keeps its old filename and sends no parameter, an unknown family exits non-zero,
  and the benchmark auto-download forwards its family.

What `swarm champion` returns, against the live API:

| Command | UID | Family | Score |
|---|---|---|---|
| `swarm champion` (before and after) | 236 | `cf_interceptor` | 0.9992 |
| `swarm champion --family-id cf_autopilot` | 161 | `cf_autopilot` | 0.8840 |
| `swarm champion --family-id cf_search_and_rescue` | 169 | `cf_search_and_rescue` | 0.9020 |

### How was this tested?

- Commands run: both CI jobs, in the CI image, on the branch.

<details>
<summary>pytest, in ghcr.io/swarm-subnet/swarm:base</summary>

```
$ pytest tests -n4 --dist worksteal -p no:cacheprovider
1005 passed, 76 skipped in 382.70s

$ pytest miner/tests -n 0 -p no:cacheprovider
35 passed in 34.21s
```

</details>

- Result: 1005 passed and 35 passed, no failures and no errors. The four new tests are in
  `tests/test_cli.py`, which is 19 passed on its own.

Then a real download against the production API, which exercises the SHA-256 check the command runs
against the hash the backend reported:

```
$ swarm champion --family-id cf_search_and_rescue
Champion: UID 169  Score: 0.9020
Per-map:  city: 0.981, forest: 0.828, mountain: 0.744, open: 1.000, village: 0.947, warehouse: 0.908
Downloading to champion_cf_search_and_rescue_UID_169.zip ...
Saved: champion_cf_search_and_rescue_UID_169.zip (1.9 MB)
```

### Tools & Dependencies

- Tools updated: None

Miners get a flag they did not have. `swarm champion` and `swarm benchmark --model ...` behave
exactly as before, including the output filename, so nothing existing has to change.

The one difference without the flag is `swarm benchmark` with no `--model` and a non-default
`--family-id`: it now downloads that family's champion rather than the cross-family best. That is
the fix, but a miner who had built a habit around the old file landing at
`champion_UID_{uid}.zip` will see a different UID in it.

### Risk & Rollback

The flag is opt-in and the unfiltered path is byte-for-byte the code that was there before, so the
blast radius is a miner who passes `--family-id` and the benchmark auto-download. Revert the commit;
there is no state, no migration and no config behind it.

### Documentation

- [x] `docs/` updated

`docs/CLI_readme.md` gains the flag in the `swarm champion` options and an example. The
`swarm benchmark` line there claimed the auto-download fetched the cf_autopilot champion, which was
not true before this change either; it now describes what the code does.

### DIFF
--- a/docs/CLI_readme.md
+++ b/docs/CLI_readme.md
@@ -68,7 +68,7 @@ swarm benchmark --model Submission/submission.zip --seeds-per-group 1
 swarm benchmark --model Submission/submission.zip --workers 3 --relax-timeouts --rpc-verbosity low
 ```
 
-If `--model` is omitted, the current champion (the default-family / cf_autopilot champion) is downloaded (with SHA-256 verification) and benchmarked instead.
+If `--model` is omitted, the champion of the family given by `--family-id` is downloaded (with SHA-256 verification) and benchmarked instead. Without `--family-id` the backend returns the highest-scoring champion across all families.
 
 Useful options:
 
@@ -266,11 +266,15 @@ swarm champion
 
 # Save to a specific path
 swarm champion --output my_champion.zip
+
+# Download the champion of one challenge family
+swarm champion --family-id cf_search_and_rescue
 ```
 
 Options:
 
 - `--output <path>`: output file path. Defaults to `champion_UID_{uid}.zip` in the current directory.
+- `--family-id <id>`: challenge family to download the champion for. Without it the backend returns the highest-scoring champion across all families, which is rarely the one you are working on. With it, the output file defaults to `champion_{family_id}_UID_{uid}.zip` so champions from different families do not overwrite each other.
 - `--backend-url <url>`: override the backend API URL (defaults to the public API).
 
 The download includes SHA-256 integrity verification against the hash reported by the backend.
--- a/swarm/cli.py
+++ b/swarm/cli.py
@@ -403,13 +403,20 @@ def _build_benchmark_argv(args: argparse.Namespace) -> list[str]:
     return argv
 
 
-def _download_champion_model() -> Optional[Path]:
+def _champion_zip_name(uid: int, family_id: Optional[str]) -> str:
+    if family_id:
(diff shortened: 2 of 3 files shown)


robojules (approved):
LGTM

PR #120, opened 2026-09-01
Move the validator's tests, scripts and docs into a validator folder

S3333F:
## Description

The miner's files moved into `miner/` in #118. This is the other half: the validator's tests,
scripts and guides now sit under `validator/`, and `docker-compose` gains a `swarm_validator`
service so each side has a one-command test run.

155 of the 177 files are pure renames. The work is in the other 22, because a move like this breaks
things that do not fail loudly — a module imported by a name that no longer exists, a path written
relative to the file holding it, a test that keeps passing while scanning an empty directory. Each
of those was found and fixed before this was opened, and each has a test that fails if it comes back.

Fixes # (issue)

### Related Tasks

- Task ID: T3CbkdMW
- Related tasks/PRs (other repos): follows #118, which moved the miner side.

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

```
tests/                     ->  validator/tests/
scripts/                   ->  validator/scripts/      (validator/{main,update,health} flattened)
docs/validator.md          ->  validator/docs/
docs/operator_runbook.md   ->  validator/docs/
```

|---|---|---|
| Imports | `scripts.*` becomes `validator.scripts.*` in 28 places | Nobody, if it is right |
| Packaging | `validator` and `validator.scripts*` ship; `validator.tests*` does not | An installed copy |
| CI / compose | Validator job runs `validator/tests`; `swarm_validator` matches `swarm_miner` | Contributors |
| Auto-updater | Real script moves; a forwarder stays at the old path | Operators: nothing |

**Module identity.** 28 runtime bindings named the `scripts` package. Nine are `import` statements;
the other **19 are module names written as strings** — `reload_module("scripts.…")` and
`monkeypatch.setattr` targets — which do not fail when a file loads, only when that line runs, or
silently patch nothing at all. Two of the nine are lazy imports inside the CLI, at `swarm/cli.py:720`
and `:782`, so `swarm --help` reaches neither: only `swarm visualize` and `swarm video` do.
`validator/__init__.py` makes the package explicit rather than a namespace one, and setuptools
selects both `validator` and `validator.scripts*` — the narrow pattern alone selects the subpackages
but not the parent, which would ship the tree without its `__init__.py`.

**File-relative paths.** 30 expressions compute a path from their own location, and after the move
not one landed where it had. They do not share a target: most want the repository root, one wants the
`swarm-backend` checkout *beside* the repository, four deliberately want their own directory. Each
was adjusted against its own target. `validator/tests/test_repo_root_paths.py` pins every expression
rather than every file — `test_github.py` has five, and a guard asking whether *any* of them reached
the root would pass with four right and one wrong.

**Discovery.** `test_scripts_shell.py` found its subjects by globbing `scripts/`, and
`test_domain_model_naming.py` asserted `offenders == []`, which is true when it scans nothing. Both
now name what they expect to find, so losing a subject fails rather than passing quietly.

**The auto-updater.** Operators register it with PM2 by path, and PM2 keeps that path across
restarts, so moving the file would have broken the updater on the pull that delivered the move — and
again on the next release, without any restart. Everything inside the scripts is self-locating and
repairs itself on pull, so only that one registered path had to survive. It is a forwarder now;
`validator/docs/validator.md` documents both paths and the `pm2 delete/start/save` steps, so it can
be dropped once operators have moved across.

### How was this tested?

Per-test outcomes were captured on `origin/main` before the move and compared after, because a moved
test can still be collected while checking less than it did. Renames are mapped rather than counted
as losses, and the new guards are excluded so the comparison stays exact.

```
baseline (origin/main)   1051 tests
after the move           1051 tests
LOST 0    ADDED 0    OUTCOME CHANGED 0    (plus 34 new guards)
```

- Full suite, in the CI image as the runner's non-root user:

```
pytest -q -n 12 --dist worksteal -p no:cacheprovider
1076 passed, 76 skipped, 272 warnings in 285.76s (0:04:45)
```

- The installed wheel, built and imported from outside the checkout with the source tree off
  `sys.path`. This caught the one defect the suite could not see: three scripts derive
  `_REPO_ROOT = _SCRIPT_DIR.parent`, which became `validator/` and, once inserted on `sys.path`, made
  the old top-level `import scripts` resolve again from an installed copy. Fixed, and pinned by a
  test that fails if the derivation is put back.

- Each guard was checked by breaking what it protects: one of `test_github.py`'s five resolvers
  regressing fails; renaming the sibling checkout fails; deleting the forwarder fails; a syntax error
  in the forwarder fails `bash -n`.

- Every local link in the repository resolves, markdown and HTML both. The first pass checked only
  markdown, and the two most visible links on the front page are HTML anchors to the validator guide.

### Tools & Dependencies

- Tools updated: None.

No. Miners are untouched. Validator operators keep working with no action: the path they registered
with PM2 still answers, and the guide documents the new one for whenever they choose to move across.

### Database & API

### Risk & Rollback

The risk in a move is a reference left behind, and the ones that matter here are module names,
computed paths and the updater's registered path. All three have tests that fail when reintroduced.
Revert the commits to put the files back; nothing outside the repository points at the new locations,
and the old updater path is answered either way.

> [!NOTE]
> Two failures predate this branch and are deliberately left alone rather than folded into a folder
> move: `gen_family_io_tables.py --check` exits 1 because it derives `interceptor_office.md` while the
> file is `office_interceptor.md`, and `contrib/CONTRIBUTING.md` links `TESTING.md` and `DEBUGGING.md`,
> neither of which exists. Both are present on `main`.

### Documentation

- [x] `docs/` updated
- [x] README.md updated


robojules (approved):
LGTM. Good job :rocket:!

PR #121, opened 2026-09-01
Add the MIT licence header to every source file

S3333F:
## Description

A source file that leaves this repository carries no statement of who wrote it or what may be done
with it. `LICENSE` covers the repository, not the single file a miner downloads or copies into their
own project. This adds the MIT header with the Swarm copyright to the top of every source file we
own, so each file is self-describing wherever it ends up.

The header text is the one already in `swarm/__init__.py`, reproduced byte for byte with the year
updated to 2026. Nothing but comment lines is added: 381 files changed, 6435 insertions, zero
deletions.

Rebased onto `main` after #120 moved the validator's tests and scripts into `validator/`, so the
headers sit on the files at their new paths and the three files that PR added are covered too.

Fixes # (issue)

### Related Tasks

- Task ID: Z7in8gdR
- Related tasks/PRs (other repos): none yet. `swarm/submission_manifest/__init__.py` is deliberately
  left without a header here, and needs a paired change in `swarm-backend` - see Additional
  Information.

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [x] Documentation

### What does this PR change?

| Area | Files | Note |
|---|---|---|
| `validator/` | 151 | the tests, scripts and docs folder #120 created |
| `swarm/core/` | 107 | mechanical |
| `swarm/validator/` | 35 | includes the runner `Dockerfile` |
| `miner/`, `scripts/`, `neurons/` | 23 | |
| `swarm/assets/` | 2 | `interceptor_drone.urdf`, `tello.urdf` |
| everything else | 63 | model graph, benchmark, challenge families, utils, base, submission template, `.docker/Dockerfile`, `conftest.py` |

Every one of those is the same edit. What is worth a reviewer's attention is the handful of files
where the header could not simply go on line 1:

- **32 files keep their first line.** 25 Python and 7 shell scripts start with `#!`; the header goes
  underneath so they stay directly executable.
- **Both Dockerfiles keep `# syntax=docker/dockerfile:1` on line 1.** A parser directive is only
  honoured as the first line of the file. Pushed down by a comment it is silently ignored and the
  build falls back to the built-in frontend with no error, so the header sits on line 2.
- **Both URDFs keep `<?xml ... ?>` on line 1**, with an XML-comment form of the header below it.
- **Two files gain one line, not a header.** `swarm/base/neuron.py` and `swarm/utils/misc.py` carry
  `Copyright © 2023 Yuma Rao` and `Copyright © 2023 Opentensor Foundation` but assert no Swarm
  copyright. They gain a single `# Copyright © 2026 Swarm` line beneath the existing notices,
  which are untouched.
- **Thirteen empty `__init__.py` files** become the header alone, with no trailing blank line.

Four files that already assert Swarm's copyright are left byte-identical, because the goal is
already met in them and their years are accurate: `neurons/validator.py` (2024),
`swarm/__init__.py` (2025), `swarm/base/validator.py` (2025), `swarm/utils/config.py` (2025).

Deliberately excluded: `swarm/templates/README.md`, whose SHA-256 is pinned on both sides and gates
miner admission; assets and generated data; JSON, which has no comment syntax; and config files.

### How was this tested?

Every changed file was reconstructed from `HEAD~1` and compared byte for byte, so the diff is
provably comment insertions and nothing else.

- Commands run: `pytest -m "not slow and not integration" -q`
- Result: `1070 passed, 74 skipped, 24 warnings, 7 errors in 290.65s`. The 7 errors are all in
  `tests/test_packaging.py` and `miner/tests/test_packaging.py`, and they error identically on a
  clean checkout of `main` on the same machine (`ModuleNotFoundError: build`). They are unrelated to
  this change.

<details>
<summary>Structural checks, all on the committed tree</summary>

```
byte-exact reconstruction from HEAD~1 : 381/381
  330 header prepended
   36 header below a preserved first line   (32 shebangs + 2 Dockerfiles + 2 URDFs)
   13 header only, empty files
    2 single line added to an existing header
removed lines                          : 0
git diff --check                       : 0 issues
file mode changes                      : 0
files carrying the header twice        : 0

python3 -m compileall                  : clean
bash -n on all 7 shell scripts         : clean
docker build -f swarm/validator/docker/Dockerfile : exit 0, image written
docker build --check, both Dockerfiles : no warnings
pybullet.loadURDF on both URDFs        : load, 5 joints each, masses and inertia unchanged
capnp.load on agent.capnp              : Agent, Observation, ObservationEntry, Tensor unchanged
swarm/__init__.py                      : identical to main, __version__ 5.1.5.3, __spec_version__ 5015
swarm/templates/README.md              : identical to main, still matches REQUIRED_README_HASH
```

</details>

### Tools & Dependencies

- Tools updated: none

No behaviour changes for miners. Validator operators see one thing, once:

> [!NOTE]
> `_calculate_docker_hash` keys the runner image on the contents of `swarm/`, so these headers change
> the cache key and the next validator run rebuilds the runner image once. Measured with the repo's
> own algorithm over its 259 inputs: `553eabf8ca765896` before, `90667ff05b48c38b` after. The rebuild
> was run here and completed successfully.

The diff touches `reward.py`, `koth.py`, `weight_utils.py` and their tests. Comment lines only; no
scoring, weight or epoch logic is altered, and no version is bumped.

### Risk & Rollback

The risk in a change this wide is a stray edit hiding among the comments, which the byte-for-byte
reconstruction above rules out. The two real hazards, a displaced Docker parser directive and a
displaced XML declaration, are both silent failures and both are covered by explicit checks. Revert
the commit to undo; nothing else depends on it.

Everyone with an open branch should expect conflicts at the top of most files.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the change is licence metadata inside source files; no documented
  behaviour, interface or workflow moved.

### Additional Information

`swarm/submission_manifest/__init__.py` is the one source file left without a header. It mirrors
`swarm-backend/app/submission_manifest.py`, and
`tests/test_submission_manifest.py::test_backend_mirror_matches_apart_from_imports` asserts the two
stay identical once import roots are normalised. Adding the header to only this side breaks that
test, so both copies should gain it in the same change when the backend is done. The test skips in
CI, which does not check the backend out alongside; it fails on a machine that has both repositories
side by side, which is where the drift would otherwise be noticed.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,4 +1,21 @@
 # syntax=docker/dockerfile:1
+# The MIT License (MIT)
+# Copyright © 2026 Swarm
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the “Software”), to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+# the Software.
+
+# THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+# THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
+# THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER IN AN ACTION
+# OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+# DEALINGS IN THE SOFTWARE.
+
 FROM python:3.11-slim-bookworm
 
 ENV PYTHONUNBUFFERED=1 \
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,20 @@
+# The MIT License (MIT)
+# Copyright © 2026 Swarm
+
+# Permission is hereby granted, free of charge, to any person obtaining a copy of this software and associated
+# documentation files (the “Software”), to deal in the Software without restriction, including without limitation
+# the rights to use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of the Software,
+# and to permit persons to whom the Software is furnished to do so, subject to the following conditions:
+
+# The above copyright notice and this permission notice shall be included in all copies or substantial portions of
+# the Software.
+
+# THE SOFTWARE IS PROVIDED “AS IS”, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO
+# THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL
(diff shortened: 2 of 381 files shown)


robojules (approved):
LGTM

PR #122, opened 2026-09-01
Install dependencies with uv instead of pip

S3333F:
## Description

Every dependency install in the repo ran through pip. They now run through uv, which resolves and
installs the same `requirements.txt` in a fraction of the time. The install that matters is the one
inside the evaluator: it runs per submission under a 120 second budget, so a miner whose dependencies
are slow to install fails for a reason that has nothing to do with their model.

`requirements.txt` and `pyproject.toml` are untouched. This changes the installer, not how
dependencies are declared, so there is no lockfile and no project restructure.

Fixes # (issue)

### Related Tasks

- Task ID: vYob1taY
- Related tasks/PRs (other repos): swarm-backend #31 rotates the README hash this branch changes.
  The two have to go live together - see the warning below.

### Type of Change

- [ ] Bug fix
- [x] Performance
- [x] Documentation

### What does this PR change?

|---|---|---|
| Evaluator | Per-submission install runs `uv pip install --python /opt/env/bin/python`; runtime deps moved into a venv at `/opt/env` with that directory first on `PATH` | Miners: installs finish sooner, fewer timeout failures |
| Images | Both Dockerfiles take the uv binary from `ghcr.io/astral-sh/uv:0.12.8`; build cache mounts moved from pip's cache to uv's | Nobody, once rebuilt |
| Host scripts | Validator and miner `install_dependencies.sh` install uv; both `setup.sh` install with it | Operators and miners on a fresh install |
| Update path | `update_deploy.sh` bootstraps uv before it touches the working tree | Operators: none, the update is self-healing |
| Evaluator | A submission-supplied `.pip_done` is removed after extraction | Nobody; closes a pre-existing hole |
| Starter template | The README the CLI writes installs with uv, so its hash rotates | Miners publishing a new repo |
| Docs | Install commands in the CLI guide, miner guide, CONTEXT and two family guides | Anyone following them |

**The evaluator could not be a command swap.** The base image already ships the CPU torch build, and
`pip install --user` resolved against it, so a miner asking for `torch` got what was already there.
uv has no `--user`, and both alternatives that write to a separate directory build a fresh closure
that ignores the environment:

| Approach | Result on a submission asking for unpinned `torch` |
|---|---|
| `pip install --user` (before) | already satisfied, installs nothing |
| `uv pip install --target <dir>` | re-downloads torch |
| `uv pip install --prefix <dir>` | re-downloads torch |
| `uv pip install --python /opt/env/bin/python` (this PR) | already satisfied, installs nothing |

The default torch wheel on PyPI is the CUDA build. Re-downloading it would exceed both the 120 second
install budget and the 500 MB `fsize` ulimit the install container runs under, so every miner
requesting torch would have failed on infrastructure. Installing into an environment rather than
beside one is what preserves the old behaviour.

`/opt/env` is writable by group and other on `lib/python3.11/site-packages` and `bin` only, because
the install container runs as the validator host's uid, which is not knowable at build time. The
container that runs miner code is `--read-only`, so those bits are inert at evaluation time, and
`/usr/local` stays root-owned at 755.

> [!WARNING]
> Merge this together with the swarm-backend PR. Admission checks `REQUIRED_README_HASH` exactly, so
> whichever repo ships alone rejects new submissions until the other follows. Champions already
> published are unaffected either way - the previous hash stays in the backend's accepted list.

### How was this tested?

- Commands run, in an image built from this branch's `.docker/Dockerfile`:
  - `pytest -q tests -n4 --dist worksteal -p no:cacheprovider`
  - `pytest -q miner/tests -n 0 -p no:cacheprovider`
  - `bash -n` on all five modified shell scripts
- Result: 1012 passed, 76 skipped, 0 failed; 35 passed; all scripts parse.

Both images were built from this branch and exercised against the evaluator's real `docker run`
flags - same uid mapping, `--cap-drop ALL`, `--security-opt no-new-privileges`, `--pids-limit=50`,
`nofile=256:256`, `fsize=524288000` - then committed and re-run under `--read-only --tmpfs
/tmp:rw,noexec,nosuid,size=64m`.

<details>
<summary>Evaluator behaviour, measured</summary>

```
base image        python -> /opt/env/bin/python
                  torch 2.14.0+cpu, torch.version.cuda is None

requirements: torch, numpy, msgpack   (all already in the base image)
  Using Python 3.11.16 environment at: /opt/env
  Checked 3 packages in 2.03s          <- nothing downloaded

full install -> docker commit -> read-only run
  torch case              6s   (budget 120s)
  tqdm + matplotlib       4s   (budget 120s)
  both imported successfully in the committed image
```

</details>

<details>
<summary>pip vs uv, same packages, same sandbox flags, cold cache both sides</summary>

```
numpy, scipy, opencv-python-headless, msgpack, tqdm, pillow

pip install --no-cache-dir --user     28.6s
uv pip install --no-cache             7.6s
```

One measurement on one requirement set, not a general ratio.

</details>

The new `tests/test_uv_dependency_install.py` was mutation-checked: removing the `.pip_done` unlink
from `batch.py` fails it, and adding a `PATH` key to a challenge family's `docker_env` fails it.

### Tools & Dependencies

- Tools updated: pip -> uv 0.12.8 as the dependency installer. Pinned in five places - both
  `COPY --from` lines, both `install_dependencies.sh` scripts, and the `update_deploy.sh` bootstrap.
- Libraries / dependencies added or bumped (name + version): none. `requirements.txt` is unchanged.

Miners: dependency installs finish sooner, which matters against the 120 second budget. A miner
publishing a new repo gets the new starter README; already-published repos keep working. The
documented install commands change from `pip install -e .` to `uv venv`, `source .venv/bin/activate`,
`uv pip install -e .`, because uv refuses to install with no virtualenv active.

Validator operators: `install_dependencies.sh` now installs uv, and an existing machine that never
ran it has uv installed for it by `update_deploy.sh` on the next update.

`swarm/__init__.py` is deliberately left at 5.1.5.3; the version moves with the next release. Note
that until it does, `auto_update_deploy.sh` will not deliver this to validators.

### Risk & Rollback

If the evaluator install misbehaves, a per-model image is built without the dependencies it needs and
the affected miner scores zero on infrastructure rather than on merit. Revert the commit: the
evaluator Dockerfile change alters `_calculate_docker_hash`, so reverting rebuilds the base image and
purges the cached per-model images on its own, and the next evaluation reinstalls with pip.

The starter README is the part that does not revert cleanly on its own - see the warning above.

### Documentation

- [ ] README.md updated
- [x] `docs/` updated
- [ ] Not applicable (explain why):

### Additional Information

Two things left alone on purpose:

- `check_readme_matches()` in `swarm/utils/github.py` still compares against the single hash rather
  than the accepted set. Nothing in the repo calls it, and three existing tests patch that constant.
- `update_deploy.sh` still resets the working tree before it reinstalls, so an install that fails
  after the reset leaves the local version matching the remote and the watcher does not retry. That
  predates this change; the new uv bootstrap deliberately sits before the reset so it cannot make it
  worse.

CI pulls `ghcr.io/swarm-subnet/swarm:base` and no workflow in this repo builds it, so CI keeps
testing against the previous image until someone rebuilds that tag from the new `.docker/Dockerfile`.

### DIFF
--- a/.docker/Dockerfile
+++ b/.docker/Dockerfile
@@ -1,9 +1,10 @@
 # syntax=docker/dockerfile:1
 FROM python:3.11-slim-bookworm
 
+COPY --from=ghcr.io/astral-sh/uv:0.12.8 /uv /uvx /bin/
+
 ENV PYTHONUNBUFFERED=1 \
     PYTHONDONTWRITEBYTECODE=1 \
-    PIP_DISABLE_PIP_VERSION_CHECK=1 \
     PYTHONPATH=/app \
     OMP_NUM_THREADS=1 \
     OPENBLAS_NUM_THREADS=1 \
@@ -16,17 +17,17 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         libgl1 libglib2.0-0 libsm6 libxext6 libxrender1 libgomp1 \
     && rm -rf /var/lib/apt/lists/*
 
-RUN pip install --no-cache-dir --upgrade pip wheel
+RUN uv pip install --system --no-cache --upgrade wheel
 
 # CPU-only torch at the exact pin from requirements.txt, so the next
 # install treats it as already satisfied instead of pulling the CUDA wheel.
-RUN pip install --no-cache-dir torch==2.10.0 \
+RUN uv pip install --system --no-cache torch==2.10.0 \
         --index-url https://download.pytorch.org/whl/cpu
 
 COPY requirements.txt /tmp/requirements.txt
 
 # setuptools 82 drops pkg_resources.fixup_namespace_packages, which pytest's monkeypatch imports.
-RUN pip install --no-cache-dir -r /tmp/requirements.txt "setuptools<82"
+RUN uv pip install --system --no-cache -r /tmp/requirements.txt "setuptools<82"
 
 WORKDIR /app
 CMD ["pytest"]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,4 +1,6 @@
 .git
+.venv
+venv
 **/__pycache__
 **/*.pyc
 swarm/assets
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -95,7 +95,9 @@ Avoid:
 Local install:
 
 ```bash
-pip install -e .
+uv venv
+source .venv/bin/activate
+uv pip install -e .
 ```
 
 Common CLI workflow:
--- a/docs/CLI_readme.md
+++ b/docs/CLI_readme.md
@@ -7,13 +7,16 @@ Command-line interface for benchmarking, testing, and packaging drone navigation
 ## Install
 
 ```bash
-pip install -e .
+uv venv
+source .venv/bin/activate
+uv pip install -e .
 ```
 
 Or install from PyPI (the published release may lag this repo):
 
 ```bash
-pip install swarm-sotapilot
+uv tool install swarm-sotapilot
+uv tool update-shell
 ```
 
 Then use `swarm <command>` directly. Alternatively, run without installation:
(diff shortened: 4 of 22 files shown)


robojules (approved):
LGTM, with one thing left before we merge: a new image with the latest changes needs to be pushed to GitHub Packages, replacing the old one.

One thing I noticed while testing: after building the Docker image, tests start failing in miner and validator side. So a good order might be to merge this PR, update the GitHub package, and then open a follow-up PR to fix the tests. Amazing job 🦾 !

PR #124, opened 2026-09-01
Rearrange the office per episode and calibrate its telemetry to the rig

S3333F:
## Description

The office interceptor rewarded memorising the room: the current champion ships a 5 cm voxel scan of the office and a photo-recogniser trained on it, and both keep working because the furniture never moves. It also ran on telemetry calibrated against a public Tello dataset from before the real rig existed.

Two changes, on one branch. The office now ships as an empty room plus 41 pieces of furniture, each baked on its own, and a seeded layout rearranges them every episode by the rules a real office follows - so a baked map is worthless and a policy has to localise from what it senses. And the telemetry moves to the values measured on the rig in the office: a velocity estimate that drifts like optical flow, ToF noise of 0.13 m with spurious ranges, and a target whose apparent width is dealt over 12-30 cm.

The detector emulator is left as it is: the rig's detector was measured and matches the current emulation.

Version 5.1.5.4 -> 5.1.5.5 so validators pick the release up. The last digit keeps every family's scores valid; only the office champion is re-evaluated, by hand, once validators are on the new version.

Fixes # (issue)

### Related Tasks

- Task ID: xMbeRPuV

### Type of Change

- [x] Documentation

### What does this PR change?

|---|---|---|
| Map | Empty room + 41 furniture pieces placed per episode by a seeded layout; furniture keeps its true size while the room's scale still varies | Miners: a memorised map no longer lines up with the room |
| Assets | Shell and floor rebaked with no furniture shadows; `pieces/` and `fixed/` folders with per-piece bakes; the three furniture-zone meshes removed (15.6 -> 17 MB) | Nobody |
| Telemetry | Horizontal velocity carries a per-episode bias that random-walks; ToF noise 0.03 -> 0.13 m with 2% spurious ranges | Miners: dead reckoning drifts, ToF needs filtering |
| Target | Apparent width dealt over 0.12-0.30 m per episode | Miners: box width no longer maps to range |
| Docs | Family guide updated to match | Nobody |
| Version | 5.1.5.4 -> 5.1.5.5 | Validator operators: auto-update picks it up; every score stays valid |

Map

- `swarm/core/maps/office/layout.py`: `office_layout(seed, scale)` is a pure function of the seed and the room scale. Wall units (shelves, cabinets, wardrobe, printer table) go back-to-wall on a wall run in their zone; the four desks stay two back-to-back pairs with their chairs and drawers, in a row along either axis; the meeting table keeps its six chairs; the lounge moves as one group against a wall; bins, lamps and stools land near a wall or a desk. The desk row and the meeting table may swap rooms. Doors keep a 1 m apron, the two columns and the radiator are obstacles, and the floor must stay one region a 0.9 m walker can cross that reaches both doors. Every footprint is axis-aligned, so overlap and clearance are box tests. An attempt with an unplaced piece or a split floor is retried on the next sub-seed; 60 of 60 seeds tried placed every piece with no overlap between groups.
- `builder.py` loads the room, the seven fixed items (partitions, wall pictures, clock) and the placed pieces as static bodies with exact mesh collision, and returns the layout in `map_info["layout"]`. Repeat-episode build cost measured on a 12-core box: 0.57 s (3.5 s on the first build in a process; PyBullet caches the mesh files).
- The family keys its flight-volume cache on the seed as well as the scale, since the furniture now moves with the seed, and keeps the layout on the env for tests.
- The clean shell and floor bakes reproduce the shipped atlases' tone (mean 166 vs 173 on the shell, 0% clipped on both); the west wing is 4% dimmer without the shelves' bounce light.

Telemetry

- `vf` and `vr` carry a bias drawn per episode from the telemetry stream (sigma 0.006 m/s per axis) that random-walks per packet (sigma 0.0005 m/s). Integrating the reported velocity now drifts and keeps drifting; before, the error was bounded by the 10 Hz sampling alone. `vd` is untouched - it does not come from optical flow.
- Downward ToF noise 0.03 -> 0.13 m. 2% of packets return a range drawn uniformly over the sensor's span instead of the true one.

Target

- Apparent width dealt uniformly over 0.12-0.30 m per episode (was 0.18 m +/-10%); aspect ratio +/-10% on top. `office_target_silhouette(seed)` is a pure function of the seed, like the airframe and target profiles.

Tests

- Map: bodies present; on seven seeds every piece is placed once, inside the room, with nothing from another group overlapping it and every chair and drawer beside its desk; the layout differs between seeds and repeats for the same seed; furniture is not scaled with the room. Two existing tests that assumed the fixed room now read the layout instead.
- Telemetry: the width band is dealt end to end; the velocity bias equals the dealt state on every packet, differs per seed and is absent on the vertical axis; ToF outliers forced on and off.

Observation layout, scoring, the detector and the contract are unchanged. Every existing submission keeps running.

### How was this tested?

- Commands run, on the branch merged with today's `main`: `pytest validator/tests -n8 --dist worksteal`, `pytest miner/tests -n 0`, `pre-commit run --all-files`
- Result: validator 1066 passed, 77 skipped; miner 46 passed; every pre-commit hook passed. The packaging tests build the wheel and pass (22 passed) once `wheel` is installed in the environment. Map file 14 passed, family file 57 passed.

Telemetry calibration against the rig, on the same 8 seeds before and after, comparing every delivered packet with the clean snapshot it was built from and integrating the reported velocity the way a miner would:

| | before | after | rig |
|---|---|---|---|
| ToF sigma (clean packets) | 0.029 m | 0.129 m | 0.13 m |
| ToF spurious ranges | 0% | 2.0% | - |
| Dead-reckoning error at 20 s | 0.21 m | 0.27 m | 0.17 m |
| Dead-reckoning error at 60 s | 0.24 m | 0.61 m | - |

The 20 s figure is the 10 Hz sampling floor (0.21 m, unchanged) plus a bias term of ~0.17 m, added in quadrature - the bias term alone matches the rig. Before this change the error stopped growing after 10 s; now it is still growing at 60 s.

Current office champion (uid 175, the exact artifact the validators scored, hash-checked) flown on the same 50 seeds against `main` and against this branch, with the camera on and the validator's scoring:

| | `main` | this branch |
|---|---|---|
| Catches | 49 / 50 | 50 / 50 |
| Mean score | 0.970 | 0.990 |
| Median time to catch | 6.5 s | 7.1 s |
| Mean time to catch | 7.7 s | 8.0 s |
| Catches under 10 s | 39 / 49 | 35 / 50 |
| Slowest catch | 30.5 s | 27.6 s |
| Collisions | 1 | 0 |

Paired per seed: median delta 0.000, worse on 6 seeds, better on 7, unchanged on 37. Fifty seeds cannot resolve a few-percent crash rate, so the same champion was then flown on 1000 seeds (0-999) against this branch alone:

| uid 175, this branch, 1000 seeds | |
|---|---|
| Catches | 962 / 1000 (96.2%) |
| Mean score | 0.946 |
| Median score | 1.000 (702 seeds at 1.0) |
| Median / mean time to catch | 7.0 s / 8.9 s |
| Misses | 35 collisions, 3 tilts |

Its live benchmark score on today's fixed office is 0.979 (validators, 1100 seeds), so the rearranged office costs it about 0.03, and nearly all of that is crashes: a 3.5% collision rate against ~1% today. The champion catches by sight once its memorised map stops matching, but it also steers by that map, and furniture that is no longer where the map says it is gets hit. Per 100-seed block the mean ranges 0.92-0.98, so the 1000-seed figure is good to about +/-0.01. What the change removes is the value of shipping a map of the room; a policy that senses the room keeps the crown.

### Tools & Dependencies

- Tools updated: none

Miners: the office is arranged differently every episode, so a map baked from one layout no longer matches the room; telemetry is noisier in the ways listed above. Nothing changes shape in the observation.

- [x] Scoring, weights or epoch behavior changes - no formula changes, but the challenge does, so the office champion is re-evaluated once on the new room; nothing else is re-scored

### Risk & Rollback

A layout rule that is off makes an episode look unusual, not broken: every piece is checked against the room, the doors and every other piece before it is placed, and an attempt that fails is retried; the last resort keeps a piece at its real-office position on free floor. If a telemetry constant is off, the family is harder or easier than the rig. Revert the commits; the assets folder returns with them.

### Documentation

- [x] `docs/` updated

### Additional Information

One number is an estimate until the raw rig recordings arrive: the ToF outlier rate (2%). Everything else is a measured value.


robojules (approved):
LGTM

PR #129, opened 2026-09-07
Move the miner model cache under swarm/state

S3333F:
## Description

The miner model cache was a relative path, `miner_models_v2`, created wherever the validator happened to run from. On every validator that is the repo checkout, so the cache shows up as an untracked folder in `git status` and goes into the Docker build context of every image build (1.4 GB on the owner validator). It now lives under `swarm/state/miner_models`, next to the rest of the runtime state, which git and Docker already ignore. The old folder is adopted in place on the first start, so no validator re-downloads its models or rebuilds its evaluation images.

Fixes # (issue)

### Related Tasks

- Task ID: pdYzDQ4c

### Type of Change

### What does this PR change?

- `MODEL_DIR` is anchored to the package path instead of the current working directory. Every reader (fetch, forensics, blacklist, image cleanup, doctor) goes through the constant, so nothing else moves.
- `ensure_model_dir()` moves the old `miner_models_v2` folder into the new place when the new one is missing or empty. An empty new folder still adopts, because `swarm doctor` and the troubleshooting docs both create it. A symlinked old folder becomes a symlink to the same disk.
- The validator calls it at startup before the Docker evaluator's cleanup, which prunes any model image whose zip is not in the cache. Adopting first is what keeps the images.
- Docs and the fake-model docstring point at the new path.

### How was this tested?

- Commands run: `pytest -m "not slow and not integration"` and the new `validator/tests/test_model_dir.py` (legacy folder adopted, empty new folder still adopts, populated new folder left alone, symlinked legacy folder keeps pointing at its disk, path is absolute under `swarm/state`).
- Result: 1099 passed, 76 skipped. The 9 packaging errors and the 1 runtime-caps failure reproduce identically on `main` in the same environment (the packaging fixtures need Python 3.11's `tomllib`, this venv is 3.10).

### Tools & Dependencies

- Tools updated: none

Validator operators: the cache moves from `<repo>/miner_models_v2` to `<repo>/swarm/state/miner_models` on the first start after the update, automatically. No action needed. Anyone who scripted against the old path has to update it.

### Risk & Rollback

If adoption fails on a host, the validator starts with an empty cache and re-downloads its models, which is slow but not incorrect. Revert the commit and the old relative path is back; a moved folder can be moved back by hand.

### Documentation

- [x] `docs/` updated

### DIFF
--- a/neurons/validator.py
+++ b/neurons/validator.py
@@ -41,6 +41,7 @@ import swarm
 from swarm.base.validator import BaseValidatorNeuron
 from swarm.validator.docker.docker_evaluator import DockerSecureEvaluator
 from swarm.validator.forward import forward
+from swarm.validator.utils_parts.model_fetch import ensure_model_dir
 
 
 # ───────────────────────────────────────────────────────────────────────────
@@ -232,6 +233,7 @@ class Validator(BaseValidatorNeuron):
                 )
         except Exception:
             pass
+        ensure_model_dir()
         self.docker_evaluator = DockerSecureEvaluator()
         if DockerSecureEvaluator._base_ready:
             bt.logging.info("✅ Docker evaluator ready")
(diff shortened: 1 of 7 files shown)


robojules (approved):
LGTM

PR #123, opened 2026-09-01
Add code linters and fix what they find

S3333F:
## Description

Nothing in this repository checked code style, and the one config that looked like it did — `.flake8` — ignored `E501`, `E402` and `W291` among others, excluded two files that no longer exist, and was never wired into CI. Five linters now run from a single `pre-commit` config, and a new `Lint` workflow runs the same hook set on every pull request, so a local commit and CI cannot disagree about what is acceptable.

Landing them means fixing what they find: 142 files were corrected automatically, and 112 findings needed a decision and were fixed by hand.

Fixes # (issue)

### Related Tasks

- Task ID: zfTRm1Jz
- Related tasks/PRs (other repos): swarm-backend rolls `REQUIRED_README_HASH` to match the template this PR relints. Both have to land, and the backend has to be deployed, before a miner can submit against the new template.

### Type of Change

- [ ] Bug fix
- [ ] New feature or capability
- [x] Refactor / cleanup (no behavior change)
- [ ] Performance
- [x] Documentation
- [x] Tooling, CI, or infrastructure

### What does this PR change?

**Configuration** (`d9f41cd`, `7c6c34e`)

- `.pre-commit-config.yaml` runs ruff, codespell, markdownlint-cli2, shellcheck, an XML parse of the three `.urdf` files, and pre-commit's own hygiene hooks, all at pinned revisions.
- `[tool.ruff]` selects `E4,E7,E9,F,I`. The `*_parts/` packages share imports through a `_shared.py` whose `__all__` is built at runtime; ruff cannot see that, so `F401/F403/F405` are ignored per-file there. Without it `ruff --fix` empties those hubs — 50 files, 438 deleted imports, and every consumer breaks at first attribute access.
- `swarm/submission_manifest/__init__.py` ignores `I001`. It mirrors a backend file line for line, and because the import roots differ (`swarm.core.submission_policy` against `app.submission_policy`) correct alphabetical order in the two repositories is opposite, so any sorting here desynchronises the mirror permanently.
- `[tool.codespell]` ignores `froms`, `statics` and `oppen`. All three are real identifiers; `oppen` is the filename `swarm/assets/maps/other_sources/vehicles/oppen door truck.obj`.
- Exclusions are per-hook rather than global, so `check-added-large-files` still sees the asset directories.
- `.flake8` deleted.

**Automatic fixes** (`aa35032`) — 142 files, no hand edits. Import ordering and unused imports from ruff, markdown structure from markdownlint, trailing whitespace and missing final newlines.

[...]

PR #130, opened 2026-09-07
Rebuild and push the test image when the Dockerfile changes

S3333F:
## Description

The test jobs pull a prebuilt image from GitHub Packages, and until now that image was only rebuilt by hand. When the Dockerfile changed, the tests kept running on whichever image was pushed last until someone remembered to rebuild it. A new `image` job now runs first: if `.docker/Dockerfile` changed in the commit under test, it builds the image and pushes it as `swarm:base`; if not, it does nothing. The miner and validator jobs wait for it, so a Dockerfile change is tested on the image it describes in the same run.

Fixes # (issue)

### Related Tasks

- Task ID: 3Co4YKGY

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

- A new `image` job in `tests.yml` compares the commit under test with its base and rebuilds and pushes the image only when `.docker/Dockerfile` differs. The base is the merge parent on a pull request, `event.before` on a push, and the merge group's base on a merge queue run.
- When the base cannot be determined (a first push, a force push) the job rebuilds rather than skips, so an uncertain case costs minutes instead of a stale image.
- The job carries `packages: write` on its own; the workflow default stays read-only, so the miner and validator jobs are unchanged apart from `needs: image`.
- The push happens on pull requests too, not only on `main`, so the test jobs in the same run pick up the new image. A pull request that edits the Dockerfile therefore updates the shared `swarm:base` tag before it merges.

> [!NOTE]
> The push uses the workflow's own token, which works only if the package on GitHub grants this repository write access. This pull request does not touch the Dockerfile, so its own run skips the push; the first pull request that does change it is the one that exercises the path.

### How was this tested?

- Commands run: the check step's script extracted from the workflow and run against eight cases from the repository's history (the merge commits of #122 and #123, pushes across and not across the `uv` change, a zero `before`, a merge group base, an unknown event); `docker build -f .docker/Dockerfile -t swarm:base .` with the workflow's exact command; then `pytest -v miner/tests -n 0 -p no:cacheprovider` and `pytest -v validator/tests -n4 --dist worksteal -p no:cacheprovider` inside that image, with the same `docker run` flags as the jobs; `pre-commit run --files .github/workflows/tests.yml`.
- Result: all eight cases classified correctly; the image builds (2.41 GB); miner 47 passed, validator 1057 passed and 77 skipped; the linters pass. The push itself could not be run locally.

### Tools & Dependencies

- Tools updated: none

### Risk & Rollback

If the change check is wrong in the "unchanged" direction, the tests run on an old image, which is exactly today's behaviour. If the push is refused, the `image` job fails and the test jobs do not run, so a broken image never reaches them silently. Revert the commit to go back to pulling the image as pushed by hand.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the workflow file carries the explanation and nothing in `docs/` describes the image push.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,7 +23,7 @@ env:
 jobs:
   # Rebuilds the test image when the Dockerfile changed, so the jobs below run
   # on the image this commit describes rather than whichever one was pushed last.
-  image:
+  push-docker-image:
     runs-on: ubuntu-latest
     timeout-minutes: 30
     permissions:
@@ -72,7 +72,7 @@ jobs:
   # change is confined to one of them is a claim the run would then not be
   # checking, and the two are less separate than the folders suggest.
   miner:
-    needs: image
+    needs: push-docker-image
     runs-on: ubuntu-latest
     timeout-minutes: 15
     steps:
@@ -107,7 +107,7 @@ jobs:
         # than running them in order, and the names come out interleaved.
 
   validator:
-    needs: image
+    needs: push-docker-image
     runs-on: ubuntu-latest
     timeout-minutes: 30
     steps:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,10 +21,58 @@ env:
   TEST_IMAGE: ghcr.io/swarm-subnet/swarm:base
 
 jobs:
+  # Rebuilds the test image when the Dockerfile changed, so the jobs below run
+  # on the image this commit describes rather than whichever one was pushed last.
+  image:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    permissions:
+      contents: read
+      packages: write
+    steps:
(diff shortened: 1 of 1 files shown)


robojules (approved):
LGTM

.github/workflows/tests.yml
```diff
@@ -21,10 +21,58 @@ env:
   TEST_IMAGE: ghcr.io/swarm-subnet/swarm:base
 
 jobs:
+  # Rebuilds the test image when the Dockerfile changed, so the jobs below run
+  # on the image this commit describes rather than whichever one was pushed last.
+  image:
```
robojules: Can you change the service name to` push-docker-image`? Thanks

robojules (dismissed):
PTAL. Thanks 🙏

PR #131, opened 2026-09-07
Skip the camera render in office tests that never read it

S3333F:
## Description

The full suite took 327 s on an 8-core machine, and one office test set the floor on its own: 90% of every office env step was pybullet rendering a 256x256 camera frame in software, and most office tests step hundreds of times without ever reading that frame. The shared office env now skips the render for tests that only read state, and the packaging wheel is built once per run instead of once per xdist worker that asks for it. Same assertions, same seeds, same values.

Fixes # (issue)

### Related Tasks

- Task ID: MvXdoSwg
- Related tasks/PRs (other repos): swarm-subnet/swarm-backend#34 turns on `-n auto` for the backend suite

### Type of Change

- [ ] Bug fix
- [x] Performance
- [ ] Documentation

### What does this PR change?

- `validator/tests/test_office_family.py`: the module-scoped env is now `office_world`, and `office_env` hands the same env to a test with the camera render replaced by a blank frame for that test only. The five tests that read pixels (obs shape and range, stream cadence, determinism, appearance by seed, target invisibility) keep real frames. The three tests that build their own env and step hundreds of times get the same blank frame through `_blind`.
- `conftest.py`: `wheel_contents` builds into a directory shared by every xdist worker for the session, under a `filelock` lock. The first worker builds, the others wait and read its wheel. Serial runs use the session directory directly, so a wheel never leaks across sessions.
- No change to what the office detector, telemetry, target flight or catch logic compute: the detector fires ray casts against the geometry, not the image.

> [!NOTE]
> Collection was not the cost. A clean install of `requirements.txt` ships no `dash` or `jaxtyping` pytest plugin, and collection takes 1.5 s. The 23.6 s figure came from a polluted environment, so `pytest.ini` is unchanged.

### How was this tested?

- Commands run: `pytest` (default `-n auto --dist worksteal`) on the same 8-core machine before and after, back to back; `pytest validator/tests/test_office_family.py -n0` after; `pytest miner/tests/test_packaging.py validator/tests/test_packaging.py -n 2` after.
- Result:

| Run | Before | After |
|---|---|---|
| Full suite wall time | 327 s | 84 s |
| Full suite result | 1102 passed, 76 skipped | 1105 passed, 76 skipped |
| `test_office_detector_recall_emerges` | 223 s | 13 s |
| `test_office_target_pause_prob_respected` | 102 s | under 7 s (no longer in the slowest 20) |
| Office file alone, serial | not run | 140 s, 55 passed |
| Wheel builds per run | 2 | 1 (second worker waits 23 s on the lock) |

The three extra passes before were environment failures (missing `wheel`, setuptools 82) fixed by matching the Docker image, not by this change.

<details>
<summary>Where an office step went before this change (cProfile, 200 hover steps)</summary>

```
200 steps 15.74s -> 78.7 ms/step
  14.228s  {built-in method pybullet.getCameraImage}   100 calls
   0.459s  DSLPIDControl.computeControl
   0.324s  office_interceptor.advance_world
   0.059s  pybullet.stepSimulation
```

</details>

### Tools & Dependencies

- Tools updated: None
- Libraries / dependencies added or bumped (name + version): None (`filelock` was already pinned in `requirements.txt`)

### Risk & Rollback

The fifty office tests that now use a blank frame no longer exercise the renderer in passing, so a renderer crash deep into a long hover would only be caught by the five RGB tests, which step 10 to 15 times. Revert the commit to get the old behaviour back; nothing outside the test tree changes.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): test-only change, no behaviour or interface moved.

### Additional Information

The remaining top entry is `test_office_spawns_spread_over_the_floor` at about 50 s: it builds 24 office maps at roughly 2 s each. That is map build cost, not rendering, and is out of scope here.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -25,6 +25,7 @@ import tempfile
 from pathlib import Path
 
 import pytest
+from filelock import FileLock
 
 
 def _ensure_repo_on_syspath() -> None:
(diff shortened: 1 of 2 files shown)


robojules (approved):
LGTM

PR #133, opened 2026-09-08
Keep leased seeds through a validator restart

S3333F:
## Description

A validator restarted by the auto-updater took its leased seeds down with it and then fought its own previous session on the backend. The first heartbeat of the new process is refused as a duplicate while the old session still looks alive, the process exited on that refusal, pm2 started it again, and the loop repeated for up to three minutes. Each life pulled a task and claimed seeds before its first heartbeat, then dropped them. On 2026-09-08 that produced 26 rejections across three validators in 35 minutes and seeds claimed up to six times.

The validator now settles its session before it takes any work, and hands its seeds back on the way out.

Fixes # (issue)

### Related Tasks

- Task ID: RzpmheOR
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/swarm-backend/pull/35 (depends on this one and this one on it; merge as a pair)

### Type of Change

- [x] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

- Startup: the backend client posts an idle heartbeat with an empty in-flight list before the first sync or task request. On a duplicate-session rejection it retries every 15 s for up to 4 min instead of exiting; a rejection that outlives that window still exits, since only a real second instance survives the takeover window.
- Shutdown: SIGINT and SIGTERM raise `KeyboardInterrupt` in the main thread, which posts one stand-down heartbeat with a 1.2 s budget before the process leaves. The backend returns every seed the validator held on that message. After the stand-down the client refuses any further evaluating heartbeat so a late timer tick cannot re-lease what was just handed back.
- The duplicate-instance fence is unchanged outside the startup announce.
- Three new tests: the announce waits out a fresh previous session, exits when the rejection outlives the window, and the stand-down posts the right payload and silences later progress heartbeats.

```mermaid
flowchart TD
  A[pm2 restart] --> B[SIGINT]
  B --> C[stand-down heartbeat: nothing in flight]
  C --> D[backend returns every held seed, attempt untouched]
  D --> E[new process starts]
  E --> F[idle heartbeat before any work]
  F --> G{previous session still fresh?}
  G -- yes --> H[wait 15 s, retry, up to 4 min]
  H --> F
  G -- no --> I[sync, next task, claim seeds]
```

### How was this tested?

- Commands run: `docker run ... ghcr.io/swarm-subnet/swarm:base pytest -q validator/tests -n4 --dist worksteal -p no:cacheprovider` (the CI image), then `pre-commit run --all-files`
- Result: 1076 passed, 77 skipped in 2 min 25 s; pre-commit clean.

### Tools & Dependencies

- Tools updated: None

Validator operators: a restart now logs "Previous validator session still fresh on the backend; retrying startup heartbeat" for up to three minutes instead of exiting and being restarted by pm2. No config change. A genuine second instance on the same hotkey still exits, after the window.

### Risk & Rollback

If the backend is unreachable at startup the announce returns a transport error and the validator continues as before; only a duplicate rejection makes it wait. The stand-down runs inside pm2's default 1.6 s kill window; if it does not land, the backend's reaper returns the seeds two minutes later. Revert the commit to restore the old startup and exit paths.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): no operator steps change; the new log line explains itself.

### Additional Information

No version bump; this ships with the next release.

### DIFF
--- a/neurons/validator.py
+++ b/neurons/validator.py
@@ -17,7 +17,9 @@
 # DEALINGS IN THE SOFTWARE.
 
 
+import asyncio
 import os
+import signal
 import sys
 import time
 from pathlib import Path
@@ -39,6 +41,7 @@ from loguru import logger
 
 import swarm
 from swarm.base.validator import BaseValidatorNeuron
+from swarm.constants import STAND_DOWN_TIMEOUT_SEC
 from swarm.validator.docker.docker_evaluator import DockerSecureEvaluator
 from swarm.validator.forward import forward
 from swarm.validator.utils_parts.model_fetch import ensure_model_dir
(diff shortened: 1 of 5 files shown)


robojules (approved):
LGTM! Good job 🚀 !

PR #134, opened 2026-09-09
Move maps, robot descriptions and textures to the swarm-worlds package

S3333F:
## Description

The map assets, the two drone URDFs and the textures the map builders apply were 300 MB of the swarm checkout and 191 MB of every wheel, and each map change went through the same history as the validator code. They now live in the public [swarm-subnet/swarm-worlds](https://github.com/swarm-subnet/swarm-worlds) repository as a small Python package, pinned to a release tag from `requirements.txt`. The simulator reaches them through `swarm_worlds.maps_dir()`, `robots_dir()` and `textures_dir()`; every relative asset name below those roots is unchanged, so the generators, the lost-person manifest and the office pieces file are untouched.

pixi is declared in `pyproject.toml` with a committed `pixi.lock`, which records the exact swarm-worlds commit behind the tag.

Fixes # (issue)

### Related Tasks

- Task ID: 4rUPlkMj
- Related tasks/PRs (other repos): [swarm-worlds v1.0.0](https://github.com/swarm-subnet/swarm-worlds/releases/tag/v1.0.0) is the release the pin points at; [swarm-worlds #2](https://github.com/swarm-subnet/swarm-worlds/pull/2) is the whole repository: the package, its README and the world pictures this README now loads.

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

|---|---|---|
| Assets | `swarm/assets/maps`, `tello/`, `tello.urdf`, `interceptor_drone.urdf` and `tao.png` are deleted; the files are byte-identical in swarm-worlds v1.0.0. `swarm/assets/map_images` is deleted too and the README loads those pictures from the swarm-worlds repository. The logo, the drone picture and the two recordings stay | Nobody |
| Simulator | Eleven loaders derive their root from `swarm_worlds` instead of `__file__` arithmetic; a dead assets constant in the open builder goes | Nobody: render identity matches main |
| Dependencies | `swarm-worlds @ git+https://github.com/swarm-subnet/swarm-worlds.git@v1.0.0` in `requirements.txt`; pixi tables in `pyproject.toml`; `pixi.lock` | Validator operators: the next auto-update installs one more package |
| CI | The GHCR image rebuild also triggers on a `requirements.txt` change, and the image opens the worlds package for writing because the simulator caches split meshes next to the assets and CI runs as the host uid | Nobody |
| Tests | `test_worlds_pin.py` keeps `requirements.txt`, `pixi.lock` and the installed package on the same tag; the resolver registry drops the two files that no longer walk up from `__file__` | Nobody |
| Docs | The README links the swarm-worlds repository from the worlds section; context, architecture, CLI and miner guides name the package and the pixi commands | Miners reading the install section |

- pixi resolves the dependencies through the editable install, so `requirements.txt` stays the single list. Declaring the git pin in both places makes uv refuse to resolve ("conflicting URLs"), which is why the pixi table holds only the project and the CPU torch index.
- `swarm-bullet3` ships `manylinux_2_34` wheels only; the pixi platform entry declares glibc 2.34 so the resolver finds them.
- The validator auto-updater and both `setup.sh` scripts need no change: `uv pip install -e .` and `uv pip install -r requirements.txt` both install the git pin.

### How was this tested?

- Commands run (in the rebuilt test image, as the host uid, the way CI runs them):
  - `pytest -v miner/tests -n 0 -p no:cacheprovider`: 47 passed
  - `pytest -v validator/tests -n8 --dist worksteal -p no:cacheprovider`: 1075 passed, 77 skipped
  - `validator/scripts/verify_render_identity.py --json` on this branch and on `main`, seven family configurations: the two JSON files are identical (`cmp` exit 0)
  - `pre-commit run --all-files`: clean
- Migration check: sha256 of all 1,720 files before the move and inside the built swarm-worlds wheel: zero differences.
- Fresh installs on a clean machine: the `setup.sh` sequence (`uv pip install -r requirements.txt` then `-e . --no-deps`), the auto-updater's `uv pip install -e .`, and `pixi install --locked` all end with `swarm_worlds` 1.0.0 importable and `swarm doctor` all OK; `pixi install --locked` a second time reports the lock still valid.
- Wheel size: `uv build --wheel` gives 191.0 MB on `main`, 17.7 MB on this branch.

### Tools & Dependencies

- Tools updated: pixi manifest and lock added (pixi 0.80.0 used to generate the lock)
- Libraries / dependencies added or bumped (name + version): swarm-worlds v1.0.0 (new, from git)

Validator operators: the next auto-update clones swarm-worlds from GitHub (about 10 s, 250 MB installed) during its usual `uv pip install -e .`. Nothing to configure. Miners on an existing checkout: run `uv pip install -r requirements.txt` once after pulling, or the map builders fail on `import swarm_worlds`.

### Risk & Rollback

If a validator host cannot reach GitHub during the update, `uv pip install -e .` fails and pm2 restarts the old code, which still has its assets. The README pictures show as broken until swarm-worlds #2 merges, so that one goes first. Rollback is a revert of this branch: the assets come back from history and the pin disappears.

### Documentation

- [x] README.md updated
- [x] `docs/` updated


robojules (approved):
LGTM, good work 🚀

PR #138, opened 2026-09-10
Load mountain and office collision trees from a per-epoch disk cache

S3333F:
## Description

Every seed rebuilds the collision tree of every static mesh from its triangles, and every miner is flown on the same 1,100 seeds, so a validator builds the same trees once per miner. The engine can now save a built tree and load it back for the same triangles at the same scale. The mountain hills and peaks and the office pieces ask for that, and the seed manager gives the engine a folder per epoch and removes the older epochs' folders, because the seeds change with the epoch and their files never hit again. On a wheel without the flag the maps build exactly as before.

Fixes # (issue)

### Related Tasks

- Task ID: WhMmlUkw
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/bullet3-swarmfork/pull/7 adds the flag and the load path; this PR is a no-op until that wheel is pinned

### Type of Change

- [ ] Bug fix
- [x] Performance
- [ ] Documentation

### What does this PR change?

- World generation: the mountain shape cache and the office collision helper add `GEOM_CONCAVE_BVH_CACHE` to their concave shapes, through `getattr` so an older wheel ignores it.
- Validator: the seed manager exports `SWARM_BVH_CACHE_DIR` pointing at `state/bvh_cache/epoch_<n>` whenever it loads or generates an epoch's seeds, creating that folder and deleting every other one under `state/bvh_cache`. The host workers inherit the variable.
- Docstrings: every function in the three touched modules now has a one-line docstring, in a separate commit so the cache change reads on its own.
- Tests: `validator/tests/test_bvh_cache.py` checks that a hill and an office piece write one tree file, load it untouched and give identical ray hits and AABB, that no folder means no files, and that the seed manager owns the epoch folder on boot and on realignment. The engine checks skip on a wheel without the flag.

### How was this tested?

- Commands run, on the 12-core rendering box with the branch wheel from the fork PR:
  - `pytest -q validator/tests -n4 --dist worksteal`: 1100 passed, 77 skipped, 0 failed, on the final branch.
  - `pre-commit run --all-files`: all hooks passed.
  - `scripts/verify_render_identity.py`, 7 scenes, on the master wheel and on the branch wheel with the cache empty, filled and switched off: all orbit and episode hashes identical.
  - Per-seed world build timed with `createCollisionShape` attributed, second build of the seed in one process, master wheel against the branch wheel loading from the folder:

| Map, seed | Collision shapes, before | Collision shapes, after | World build, before | World build, after |
|---|---|---|---|---|
| mountain 102 | 2,657 ms | 908 ms | 5,225 ms | 3,376 ms |
| mountain 1302 | 2,420 ms | 850 ms | 4,853 ms | 2,876 ms |
| mountain 2302 | 2,844 ms | 1,100 ms | 6,056 ms | 3,408 ms |
| office 100 | 247 ms | 91 ms | 961 ms | 607 ms |
| office 107 | 195 ms | 87 ms | 736 ms | 659 ms |
| office 114 | 209 ms | 81 ms | 872 ms | 650 ms |

- Result: about 1.7 s less per mountain seed and 0.1 to 0.2 s per office seed from the second miner on, identical observations. Six seeds left 334 files and 137 MB in the folder; the office writes one set of files per seed because the room is rescaled per seed, so the folder grows over an epoch and is dropped at the next one.

### Tools & Dependencies

- Tools updated: None
- Libraries / dependencies added or bumped (name + version): None here; the effect starts when `swarm-bullet3` is pinned to the wheel carrying the flag.

Validator operators get a `state/bvh_cache` folder that grows over the epoch and is cleared at the epoch change. Miners see nothing.

### Risk & Rollback

If the engine's key check ever let a wrong tree through, contacts would differ from a fresh build; the identity run and the tests are the guard, and the file is keyed on the scaled triangles. Rollback is reverting this commit or removing the environment variable, after which every shape builds as before.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the behaviour is internal to the validator and explained in the code.

### DIFF
--- a/swarm/core/maps/office/builder.py
+++ b/swarm/core/maps/office/builder.py
@@ -65,6 +65,7 @@ _WINDOW_PLUG_CENTER: Tuple[float, float, float] = (17.99, 3.8, 1.65)
 _WINDOW_PLUG_HALF_EXTENTS: Tuple[float, float, float] = (0.02, 1.26, 0.76)
 
 def _asset(name: str) -> str:
+    """Absolute path of an office asset file, failing loudly when it is missing."""
     path = os.path.join(_OFFICE_ASSET_DIR, name)
     if not os.path.exists(path):
         raise FileNotFoundError(f"office map asset missing: {path}")
@@ -75,11 +76,13 @@ _PIECE_FILES = {item["id"]: item["file"] for item in office_pieces()["pieces"]}
 
 
 def _visual_shape(cli: int, obj_path: str, scale) -> int:
+    """A visual mesh shape for one office OBJ at the episode's scale."""
     return p.createVisualShape(p.GEOM_MESH, fileName=obj_path,
                                meshScale=list(scale), physicsClientId=cli)
 
 
 def _collision_shape(cli: int, obj_path: str, scale) -> int:
+    """A concave collision shape for one office OBJ at the episode's scale, tree cached on disk."""
     flags = p.GEOM_FORCE_CONCAVE_TRIMESH if hasattr(p, "GEOM_FORCE_CONCAVE_TRIMESH") else 0
     flags |= getattr(p, "GEOM_CONCAVE_BVH_CACHE", 0)
     return p.createCollisionShape(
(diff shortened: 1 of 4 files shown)


robojules (approved):
LGTM

PR #139, opened 2026-09-10
Seeded daylight sun for families that opt in

S3333F:
## Description

The seeded light direction today is a point on a circle, so half of the seeds put the light below the ground, and nothing scored reads it: depth families ignore light and the office family draws its own. A family that opts in now gets one sun per seed from a real sun arc, sunrise to sunset: the seed picks the hour and the heading, the hour sets the sun's height, and the height sets its colour and strength. Same seed, same sun on every validator; different seeds, different light. Nothing changes for a family that does not opt in.

Fixes # (issue)

### Related Tasks

- Task ID: Tg2EJyVd
- Related tasks/PRs (other repos): none. The shadow map (card 9) and the sky (card 11) will read the same per-seed sun.

### Type of Change

- [ ] Bug fix
- [ ] Performance
- [ ] Documentation

### What does this PR change?

- Simulator: a new `swarm.core.daylight` module. `seeded_sun(seed)` draws an hour between sunrise and sunset and a heading from its own random stream, puts the sun on a 40° latitude arc (noon peak 60°, never below 3°), reads the colour off a five-row table (orange at the horizon, white above 50°), and thins the sun and dims the ambient toward the horizon. Every output is rounded to fixed decimals.
- Family switch: `ChallengeFamilyRuntime.seeded_sun`, default `False`. When a family sets it, the env applies the seed's sun at construction and the two colour render calls (office observation, SAR on-demand RGB) pass the sun's colour, ambient and diffuse to the renderer. When it is `False` the calls carry exactly today's arguments.
- Constants: the arc, the minimum elevation, the extinction and the ambient range, next to the existing light flag.
- Tests: `validator/tests/test_daylight.py`, seven tests: same seed same sun, five seeds five suns, 500 seeds never below the horizon, low sun warm and weak and high sun white and strong, the render arguments, the env wiring, and every existing family still off.

### How was this tested?

- Commands run: `pytest -q validator/tests -n4 --dist worksteal` and `pytest -q miner/tests -n 0` on the branch; `validator/scripts/verify_render_identity.py` plus a colour and segmentation orbit hash on `main` and on the branch, same published wheel `swarm-bullet3 2.0.0.3`, one render thread; a 256 px colour frame of the solar-park study pieces timed with and without the sun, two render threads, median of 30.
- Result: validator 1103 passed, 77 skipped; miner 47 passed. Identity: 9 of 9 depth and episode hashes identical, 7 of 7 colour hashes identical. Frame time 69.3 ms today against 67.8 ms with the sun, so no cost: the sun is three more arguments to a call that already takes them.
- Sheet: the solar-park pieces rendered under seeds 25, 96, 57, 65 and 116 (dawn, noon, afternoon, evening, dusk) through the engine, and the same five suns through the study's ray-cast prototype with shadows. Both sheets are attached to the task card; they were briefly committed under `docs/img/` and removed again, so the branch carries no images.

### Tools & Dependencies

- Tools updated: None

No. No family opts in yet, and the identity hashes above show today's families render the same bytes.

### Risk & Rollback

Low risk. Only a family that sets `seeded_sun = True` is affected, and none does. Revert the commit to undo.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the module docstring and the constants describe the model; no family uses it yet, so no family guide changes.

### Additional Information

Sun shadows are not part of this change: the renderer's shadow pass aims at the world origin and does not cover a map. The shadow map card computes shadows from this sun direction; the sky card colours the background from the same time of day. Night is not sampled: the arc stops at 3° above the horizon, and the day and night card decides what night looks like.


docs/img/seeded_sun_engine.png
robojules, line 1: Do we need the .png? Otherwise maybe we can delete it

robojules (dismissed):
PTAL

PR #140, opened 2026-09-10
Per-family sky colours for the colour cameras, white background by default

S3333F:
## Description

The renderer can now paint a sky where a colour camera sees nothing, instead of white. A family opts in by returning a horizon and a zenith colour from `sky_colors(task)` on its runtime; the env passes them to every colour render it makes. No family returns one yet, so every existing family renders exactly the pixels it renders today.

Drone colour camera, white clear against a sky

Fixes # (issue)

### Related Tasks

- Task ID: 8wpOrkRT
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/bullet3-swarmfork/pull/10 (adds the two camera arguments; this PR passes nothing until a family opts in, so it is safe on the current wheel)

### Type of Change

- [ ] Bug fix

### What does this PR change?

- `ChallengeFamilyRuntime.sky_colors(task)`: the opt-in hook, returns `None` by default. A family overrides it to return `(horizon, zenith)` RGB triples in 0..1; a seed-dependent sky is the same override reading the task.
- `MovingDroneAviary` reads the hook once at construction and adds `skyHorizonColor` and `skyZenithColor` to its three colour renders, the office observation, the SAR on-demand RGB and the recorded video frame, only when the family returned a sky. Depth renders are untouched.
- `validator/tests/test_sky_colors.py`: every registered family returns no sky, the base runtime returns no sky, and the camera kwargs are empty or exactly the two colours.

### How was this tested?

- Commands run: `pytest -v validator/tests -n4 --dist worksteal -p no:cacheprovider` in a fresh venv from `requirements.txt`, on the sky wheel of the fork.
- Result: 1100 passed, 77 skipped.
- Render identity with this branch checked out, master wheel against sky wheel, `validator/scripts/verify_render_identity.py` plus the same orbit hashing colour and segmentation: 7 of 7 scenes identical.

### Tools & Dependencies

- Tools updated: none
- Libraries / dependencies added or bumped: none. A family that opts in needs the fork wheel that carries the two arguments; on the current pin they do not exist and the render call would raise, which is why no family opts in here.

None. No family passes a sky, so miners and validators see the same frames.

### Risk & Rollback

Low risk: one attribute read at env construction and an empty dict merged into three camera calls. Revert the commit.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): nothing a miner or operator does changes; the hook's docstring says what a family returns.

### DIFF
--- a/swarm/challenge_families/base.py
+++ b/swarm/challenge_families/base.py
@@ -198,6 +198,12 @@ class ChallengeFamilyRuntime:
         _ = task
         return {}
 
+    def sky_colors(self, task: Any) -> tuple[tuple[float, float, float], tuple[float, float, float]] | None:
+        """Horizon and zenith colours painted where a colour camera sees nothing, or None
+        for the plain white background every family renders today."""
+        _ = task
+        return None
+
     def observation_interface_version(self, task: Any) -> str:
         _ = task
         return get_supported_interface_versions(self.family_id)[0]
--- a/swarm/core/moving_drone.py
+++ b/swarm/core/moving_drone.py
@@ -207,6 +207,7 @@ class MovingDroneAviary(BaseRLAviary):
             self.GOAL_POS   = np.asarray(task.goal, dtype=float)
         self.EP_LEN_SEC = float(task.horizon)
         self.family_runtime = runtime_family_for_task(task)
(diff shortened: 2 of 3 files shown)


robojules (approved):
LGTM

PR #142, opened 2026-09-10
Per-family depth render backend with the ray caster as an opt-in

S3333F:
## Description

Depth frames on every family come from TinyRenderer. The swarm-bullet3 wheel now carries a second depth backend, a ray caster whose cost follows the pixel count instead of the map size, behind a render flag. Each family runtime declares which backend its cameras use; every family stays on TinyRenderer, and the office family, which observes colour, cannot leave it. Flipping a family is one line on its runtime class, and a version bump, because its pixels change.

Fixes # (issue)

### Related Tasks

- Task ID: wwogfkZU
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/bullet3-swarmfork/pull/11 (the backend and the flag; this branch runs on the current wheel and raises a clear error only if a family asks for the ray caster on a wheel without it)

### Type of Change

### What does this PR change?

- `ChallengeFamilyRuntime.render_backend`, default `"tiny"`. `MovingDroneAviary` reads it once at construction, `SWARM_RENDER_BACKEND` overrides it for experiments, and `"raycast"` adds `ER_SWARM_RAYCAST` to the drone camera flags on both the single and the batched depth call. The office family is pinned to TinyRenderer regardless.
- Tests: every family defaults to TinyRenderer; on the new wheel, agreement with TinyRenderer on primitives, an OBJ mesh, a back-facing quad and the Crazyflie body from two viewpoints, matching segmentation ids, a teleported body, identical frames across 1, 2 and 4 threads in separate processes, the switch on an autopilot episode, and the office pin.

### How was this tested?

- Commands run, on a 12-core AMD EPYC box with the 2.0.0.5 wheel from the fork branch: `pytest validator/tests/test_render_backend.py` and `pytest validator/tests -n4 --dist worksteal`; `verify_render_identity.py` with and without `SWARM_RENDER_BACKEND=raycast`, at 1 and 2 threads, and once more on a second box with the same wheel file.
- Result: 9 of 9 backend tests pass; validator suite 1108 passed, 77 skipped, 1 error in `test_packaging.py` from a venv without the `wheel` package; that file passes after installing it (17 passed). With the switch off, all 7 scenes hash identically to `main` on the current wheel. With the switch on, the 6 depth families produce different episode hashes, identical across the two machines and both thread counts, and the office scene is unchanged. Depth and timing per map are in the fork PR.

### Tools & Dependencies

- Tools updated: none
- Libraries / dependencies added or bumped: none in this branch; `render_backend = "raycast"` on a family needs swarm-bullet3 2.0.0.5 or later.

None. No family changes backend here, so miners and validators see the same frames.

### Risk & Rollback

Low risk: the only new code on the default path reads one class attribute and adds a zero flag. Revert the commit to remove the switch.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the attribute is documented on the base class; family authors opt in there.

### DIFF
--- a/swarm/challenge_families/base.py
+++ b/swarm/challenge_families/base.py
@@ -172,6 +172,8 @@ class ChallengeFamilyRuntimeProfile:
 class ChallengeFamilyRuntime:
     family_id: str
     runtime_supported: bool = True
+    # Depth camera backend: "tiny" is TinyRenderer, "raycast" the Embree ray caster in the same wheel.
+    render_backend: str = "tiny"
 
     def screening_policy(self) -> Dict[str, Any]:
         return get_family_screening_policy(self.family_id)
--- a/swarm/core/moving_drone.py
+++ b/swarm/core/moving_drone.py
@@ -324,6 +324,14 @@ class MovingDroneAviary(BaseRLAviary):
             os.environ.get("SWARM_BATCH_DEPTH", "1") != "0"
             and hasattr(p, "getDepthImagesBatch")
         )
+        backend = os.environ.get("SWARM_RENDER_BACKEND") or self.family_runtime.render_backend
+        # The office family observes colour, which only TinyRenderer shades, so it stays there.
+        self._raycast_enabled = backend == "raycast" and not self._office_rc_enabled
+        self._render_flags = 0
+        if self._raycast_enabled:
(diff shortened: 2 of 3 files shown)


robojules (approved):
LGTM

PR #143, opened 2026-09-10
Start the next seed's container while the current seed flies

S3333F:
## Description

Every seed runs the miner in a fresh container, and the whole container start (workspace, `docker run`, network lockdown, model load) sat in series before each flight: 15 to 23 s per seed on the fork, most of it the model load. The container depends only on the miner and the family profile, not on the seed, so a worker now starts the next seed's container while the current seed flies and adopts it when the next seed arrives. One clean container per seed stays the rule and nothing is shared between the two.

Fixes # (issue)

### Related Tasks

- Task ID: trLvmT2p

### Type of Change

- [ ] Bug fix

### What does this PR change?

- Validator and benchmark workers: the container start is split from the flight. Once a seed's container serves, the worker starts a spare for the same miner in a background thread; the next request adopts it, or discards it when the miner, profile or image differ, when the spare failed, or when the worker stops. A failed spare leaves nothing behind and the seed starts its own container, as before.
- The spare starts at the lowest CPU weight (`--cpu-shares=2`) and under a `--cpus=0.5` quota, so it takes only what the flying container leaves on the shared cpuset. Both are lifted with `docker update` before the spare is adopted; if that fails the spare is dropped and a fresh container is started.
- Both dispatchers mark each request with whether more seeds follow, so the last seed of a miner does not start a spare. In seed flow that is "pool not drained".
- The per-seed timing line now reports every start stage (setup, launch, lockdown, port up, first successful ping), whether the start was hidden behind the previous seed, and how long the seed waited for its container.
- `SWARM_DOCKER_PREWARM=0` turns the overlap off and restores the serial flow.

### How was this tested?

- Commands run: `pytest -m "not slow and not integration"` (1148 passed, 76 skipped; the two wheel-packaging tests collided under `-n auto` and pass when run alone), `pre-commit run --all-files`.
- New tests in `validator/tests/test_container_prewarm.py` drive a worker against a fake docker daemon: the second container launches mid-flight at the low weight and quota and is raised before it flies; the gap between flights drops below the model load time; a spare for the last seed, for another miner, or with a failed limit restore is killed and never flown; the switch restores the serial flow; the validator dispatcher sets the hint on every seed but the last.
- Scores and timing on a fast rented host (EPYC 7742, 128 vCPU, 6 workers, baseline model, 12 seeds, same seed file): three runs, serial, overlapped, serial again. All 12 scores identical across the three runs, every seed `seed_done`, no slow-act strike in any run.

| Run | Wall clock | Container start charged per seed | Seeds that waited 0 s for their container |
|---|---|---|---|
| serial | 237 s | 9.8 s mean (9.1 to 11.3) | 0 of 12 |
| overlapped | 225 s | 5.9 s mean; 1.0 s for a seed with a spare, 9.8 s for each worker's first seed | 6 of 12 (every seed after a worker's first) |

The serial start splits into setup 0.2 s, `docker run` 2.5 s, lockdown and gate 0.4 s, model load to first ping 5.1 s. There is no image pull: the base image is built once per version and the dependency image once per model, neither per seed.

- CPU contention on a production-shaped validator (12 vCPU, calibrated at 1.57x, 6 workers, baseline model, 12 seeds, dense act() tracing): three runs, serial, overlapped, serial again. The overlap shifts act() by +3.6 ms median per seed, worst seed +26.5 ms. Two serial runs of the same seeds shift by -8.0 ms median, worst seed -34.8 ms, so the overlap stays inside the host's own run-to-run variation. Act times there run 12 to 286 ms against a budget near 940 ms. Scores differ on 4 of 12 seeds between serial and overlapped and on 3 of 12 between the two serial runs, largest gap 0.0113 against 0.0099; every seed completed in all three runs, and the per-seed container start drops from 12.9 s and 12.1 s to 7.7 s.
- The limits were read from the live cgroups while an overlap was in progress: the flying container holds its two pinned cores at weight 100 with no quota, and the spare sits on the same two cores at weight 1 under a 0.5 core quota. The miner therefore keeps its cores whenever it wants them and the spare runs on what is left, which is why the act() shift above is in the noise.

- Timing on a slow host (rendervps, EPYC 2.0 GHz, 12 vCPU, 6 workers, 36 seeds): the serial start costs 15.3 s per seed for the baseline model and 23.4 s for a torch miner (UID 155), of which the model load is 9.7 s and 16.2 s. Overlapped, the model load drops to 1.5 s and 4.3 s and 30 of 36 baseline seeds connect on the first ping. That host runs both models at the edge of their act budget, so the score identity check ran on the fast host above.

Validator operators: no action. Each seed after the first on a worker starts flying 8 to 20 s sooner, the whole container start. During the overlap a worker briefly holds two containers, so peak RAM per worker is up to two model loads. `SWARM_DOCKER_PREWARM=0` restores the old flow.

### Risk & Rollback

If the spare competed with the flying miner for CPU, act times would rise and slow-act strikes could change a score; the low weight and the quota bound that, the live cgroup read confirms them, and the act() measurement on a production-shaped box puts the shift inside the host's own noise. Rollback: set `SWARM_DOCKER_PREWARM=0`, or revert the commit; the serial path is unchanged.

### Documentation

- [x] No docs to update: an internal validator mechanism with one env switch, described here.

### DIFF
--- a/swarm/benchmark/engine_parts/workers.py
+++ b/swarm/benchmark/engine_parts/workers.py
@@ -23,7 +23,11 @@ import signal
 
 from swarm.challenge_families import DEFAULT_RUNTIME_FAMILY_ID
 from swarm.config import HostWorkerRuntimeSettings, env_bool
-from swarm.constants import AGENT_STARTUP_WALL_SEC, MINER_COMPUTE_BUDGET_SEC
+from swarm.constants import (
+    AGENT_STARTUP_WALL_SEC,
+    MINER_COMPUTE_BUDGET_SEC,
+    WARM_CONTAINER_GRACE_SEC,
+)
 from swarm.validator.calibration.speed_factor import baseline_model_available
 from swarm.validator.docker.docker_evaluator_parts._shared import (
     _runtime_profile_from_payload,
@@ -77,7 +81,7 @@ _PR_SET_PDEATHSIG = 1
 
 # A spare start has its own deadlines; this only bounds the wait after a flight,
 # and stays well inside the parent's stall timeout for the next batch.
-_WARM_CONTAINER_WAIT_SEC = AGENT_STARTUP_WALL_SEC + 15.0
+_WARM_CONTAINER_WAIT_SEC = AGENT_STARTUP_WALL_SEC + WARM_CONTAINER_GRACE_SEC
 
 
 def _prewarm_enabled() -> bool:
--- a/swarm/benchmark/engine_parts/workers.py
+++ b/swarm/benchmark/engine_parts/workers.py
@@ -22,12 +22,19 @@ import os
 import signal
 
 from swarm.challenge_families import DEFAULT_RUNTIME_FAMILY_ID
-from swarm.config import HostWorkerRuntimeSettings
-from swarm.constants import MINER_COMPUTE_BUDGET_SEC
+from swarm.config import HostWorkerRuntimeSettings, env_bool
+from swarm.constants import AGENT_STARTUP_WALL_SEC, MINER_COMPUTE_BUDGET_SEC
 from swarm.validator.calibration.speed_factor import baseline_model_available
+from swarm.validator.docker.docker_evaluator_parts._shared import (
+    _runtime_profile_from_payload,
+)
 from swarm.validator.docker.docker_evaluator_parts.batch import (
+    WarmContainer,
+    WarmContainerStart,
     _ensure_host_speed_factor,
+    discard_warm_container,
     host_speed_factor_is_fresh,
+    warm_container_key,
 )
 
 from ._shared import (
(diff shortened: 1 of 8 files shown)


swarm/benchmark/engine_parts/workers.py
```diff
@@ -68,6 +75,15 @@
 
 _PR_SET_PDEATHSIG = 1
 
+# A spare start has its own deadlines; this only bounds the wait after a flight,
+# and stays well inside the parent's stall timeout for the next batch.
+_WARM_CONTAINER_WAIT_SEC = AGENT_STARTUP_WALL_SEC + 15.0
```
robojules: Add 15.0 as a constant | parametre to be changeable

robojules (dismissed):
PTAL

PR #145, opened 2026-09-11
Keep far bodies visible on the ray-cast path and measure mask parity

S3333F:
## Description

The distance cull hides a large body beyond 35 m by setting its colour alpha to 0, so TinyRenderer does not paint its triangles. The ray caster honours alpha 0 the same way, so on the ray-cast path the body leaves the scene as well, although a ray caster's frame cost does not grow with what is in view. On the mountain map that is 126 of the 150 cull targets gone by step 60 of an episode. A family on the ray-cast backend now keeps every body visible and only the 50 m collision cull runs. The segmentation mask on the ray-cast path was measured against TinyRenderer on the seven identity scenes and the script that does it is checked in.

Fixes # (issue)

### Related Tasks

- Task ID: U2gEF1FK
- Related tasks/PRs (other repos): builds on #142, now merged; needs swarm-bullet3 with `ER_SWARM_RAYCAST` for the ray-cast half of the tests.

### Type of Change

- [x] Bug fix

### What does this PR change?

- Distance cull: the visual hide is skipped when the env runs on the ray caster. The collision cull beyond 50 m is unchanged on both backends, and TinyRenderer keeps hiding as before.
- Tests, on a large mesh 45 m from the drone: the cull decision, which asserts the hide and the body's alpha and needs no particular wheel, and the frame it produces, which does. At 60 m both backends lose the body's collisions. Three of the four cases run in the action; the fourth is a ray-cast frame, which the action cannot render (see below).
- `validator/scripts/compare_render_masks.py`: renders the identity orbit of every `verify_render_identity` scene on both backends with the object-and-link mask on and reports, per scene, the share of pixels with the same id, split into hit-or-miss differences and different-id differences. Listed in the repo-root path test like the other scripts.

> [!NOTE]
> The action installs swarm-bullet3 2.0.0.3 from `requirements.txt`, and the ray caster exists from 2.0.0.5, which is not published. Every test behind the backend check skips there, the nine from #142 included. Publishing the wheel and moving the pin turns all of them on together; until then the cull decision above is the part of this change the action does check.

### How was this tested?

- Commands run, on a 12-core AMD EPYC box, `SWARM_RENDER_THREADS=2`, against a 2.0.0.5 wheel built at the `v3` tier from `bullet3-swarmfork` `master` at 883b032, which carries both the sky and the ray-cast backend:
  - `pytest validator/tests/test_moving_drone_cull.py` on `main` and on this branch.
  - `verify_render_identity.py` on `main` and on this branch, with and without `SWARM_RENDER_BACKEND=raycast`.
  - `compare_render_masks.py`, and a script that steps an episode on the ray caster and counts the bodies the cull has hidden.
  - `pytest validator/tests/test_render_backend.py validator/tests/test_repo_root_paths.py`, then `pytest validator/tests -n4 --dist worksteal`.
- Result: on `main` the two ray-cast cases fail, one on the cull state and one on the frame, and the three TinyRenderer cases pass; on this branch 42 passed across the three files and the validator suite is 1144 passed, 77 skipped. All 7 identity scenes hash the same on `main` and on this branch, on both backends.
- Against the 2.0.0.3 wheel the action installs, the same file is 4 passed and 1 skipped: everything except the ray-cast frame.

What the cull was doing to the ray-cast scene, and what that costs in the picture today:

| Family and map | Camera far | Cull targets | Hidden by step 60 on `main` | Depth frame vs this branch |
|---|---:|---:|---:|---|
| Autopilot, mountain t3 s103 | 30 m | 150 | 126 | identical |
| Autopilot, city t5 s105 | 30 m | 28 | 0, under the face threshold | identical |
| Autopilot, forest t6 s106 | 30 m | 0 | 0 | identical |
| SAR, mountain t2 s104 | 30 m | 1 | 0, under the face threshold | identical |
| Interceptor t2, seeds 210-214 | 110 m | 1 | 0, under the face threshold | identical |

No frame changes on any family today: the one map whose cull engages has a 30 m camera, and the cull starts at 35 m, so every body it removed was already past the far plane. The fix is what keeps that true for a map with large geometry further out, or a family whose camera reaches past 35 m.

Mask agreement, identity orbit of 40 poses per scene, `ER_SEGMENTATION_MASK_OBJECT_AND_LINKINDEX`, TinyRenderer against the ray caster:

| Scene | Pixels | Same id | Hit vs miss | Both hit, different id |
|---|---:|---:|---:|---:|
| Open, autopilot t1 s101 | 655,360 | 99.9927 % | 3 | 45 |
| Mountain, autopilot t3 s103 | 655,360 | 99.9998 % | 0 | 1 |
| City, autopilot t5 s105 | 655,360 | 99.4180 % | 42 | 3,772 |
| Forest, autopilot t6 s106 | 655,360 | 99.9973 % | 13 | 5 |
| SAR mountain t2 s104 | 2,621,440 | 99.9999 % | 1 | 2 |
| Swarm SAR t2 s1040 | 2,621,440 | 99.9938 % | 1 | 161 |
| Office t7 s7 | 2,621,440 | 99.9841 % | 0 | 417 |

On the city scene, the one with the most differing pixels, 3,550 of the 3,814 sit on an object edge, where the pixel-corner sample of each renderer lands on a different neighbour, and 212 more are two surfaces at the same depth within 1 cm. The remaining 52 pixels, 0.008 %, are neither.

Frame cost with the cull targets visible and hidden, drone camera at spawn, depth only, median of 20:

| Map | Cull targets | Target faces | TinyRenderer visible | TinyRenderer hidden | Ray cast visible | Ray cast hidden |
|---|---:|---:|---:|---:|---:|---:|
| Mountain, autopilot t3 s103 | 150 | 1,058,627 | 11.42 ms | 0.42 ms | 3.69 ms | 4.03 ms |
| City, autopilot t5 s105 | 28 | 55,077 | 4.09 ms | 2.87 ms | 5.57 ms | 6.24 ms |
| Forest, autopilot t6 s106 (no targets, so the pair is the noise floor) | 0 | 0 | 17.60 ms | 17.10 ms | 5.01 ms | 4.70 ms |

The cull exists for the TinyRenderer pair. The ray-cast pair is why the ray-cast path does not need it.

### Tools & Dependencies

- Tools updated: `validator/scripts/compare_render_masks.py` added.
- Libraries / dependencies added or bumped: none.

None. No family runs on the ray caster, and on the current maps and seeds the cull removes nothing a depth camera could see, as the table above shows.

### Risk & Rollback

Low risk: the default path is untouched, which the identity hashes show on both backends. Revert the commit to hide far bodies on the ray caster again.

### Documentation

- [ ] README.md updated
- [ ] `docs/` updated
- [x] Not applicable (explain why): the behaviour is documented in the comment next to the condition; the script carries its own usage line.

### DIFF
--- a/validator/tests/test_moving_drone_cull.py
+++ b/validator/tests/test_moving_drone_cull.py
@@ -1,3 +1,5 @@
+"""Tests for the distance cull: what it hides, what it isolates, and which backend it applies to."""
+
 from __future__ import annotations
 
 from pathlib import Path
@@ -16,6 +18,7 @@ _needs_wheel = pytest.mark.skipif(
 
 
 def test_cull_reenable_keeps_static_bodies_isolated(monkeypatch) -> None:
+    """A body whose collisions come back after the cull is still isolated from the other static bodies."""
     cli = p.connect(p.DIRECT)
     try:
         p.setAdditionalSearchPath(pybullet_data.getDataPath(), physicsClientId=cli)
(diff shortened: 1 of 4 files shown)


validator/tests/test_moving_drone_cull.py
robojules, line 1: I think this test is not being executed in the action checks? Could you please verify it? Thx!

robojules (commented):
PTAL

validator/tests/test_moving_drone_cull.py
S3333F: Verified, and half of it was not running: the TinyRenderer case ran, the ray-cast case skipped. The action installs swarm-bullet3 2.0.0.3 from requirements.txt and the ray caster exists from 2.0.0.5, which is not on PyPI, so every test behind that check skips there, the nine from #142 included.

Split so the part that guards this change no longer needs the wheel: the cull decision asserts the hide and the body's alpha, which is the mechanism the cull uses, and only the resulting frame needs the ray caster. This run executes three of the four cases, the ray-cast cull decision among them, and against the 2.0.0.3 wheel the file is 4 passed, 1 skipped. On main the two ray-cast cases fail, so the regression is covered from both sides.

Publishing the wheel and moving the pin would switch on the fourth case and the #142 tests together.

robojules (approved):
LGTM

PR #146, opened 2026-09-13
Move the workflow actions off the deprecated Node 20 runtime

S3333F:
## Description

Every run of both workflows prints the same annotation: `Node.js 20 is deprecated. The following
actions target Node.js 20 but are being forced to run on Node.js 24`. Nothing fails today, but
the runtime is being picked for us and Node 20 is being switched off. This moves all five
actions to the majors that ask for Node 24.

### Related Tasks

- Task ID: 1H1EuByM

### Type of Change

### What does this PR change?

| Workflow | Action | Before | After |
|---|---|---|---|
| Lint | `actions/checkout` | v4 | **v7** |
| Lint | `actions/setup-python` | v5 | **v7** |
| Lint | `actions/cache` | v4 (indirect) | **v6** (direct) |
| Tests | `actions/checkout` (3 jobs) | v4 | **v7** |
| Tests | `docker/login-action` (3 jobs) | v3 | **v4** |

- `actions/cache@v4` was never in this repo's files. `pre-commit/action@v3.0.1` pins it
  internally, that project is maintenance-only and has had no tag since February 2024, so the
  version cannot be reached from here. The lint job now runs the three steps that action was
  running: restore the cache, install pre-commit, run it. Same cache key, so the existing cache
  still hits.
- Every input the workflows pass (`persist-credentials`, `fetch-depth`, `python-version`,
  `registry` / `username` / `password`, `path` / `key`) is still present in the new majors,
  checked against each action's `action.yml` at the tag.
- No Python changed.

### Run it yourself

```bash
pre-commit run --all-files --show-diff-on-failure
```

That is now literally what the lint job runs, rather than a wrapper around it.

### How was this tested?

- Commands run: `pre-commit run --all-files`, and `actionlint .github/workflows/*.yml`.
- Result: **14 hooks passed in 17 s**, actionlint clean on both changed files.
- The annotation disappearing can only be seen on a real run, so the proof is this PR's own
  Lint and Tests checks.

### Risk & Rollback

Low risk: the lint job runs the same hooks from the same config, and the test jobs are unchanged
apart from the action versions. If one of the new majors misbehaves the commit reverts on its
own.

### Documentation

- [x] Not applicable: no documented behavior changed.


robojules (approved):
LGTM, good work :rocket:

PR #147, opened 2026-09-13
Add the missing docstrings across the repository

S3333F:
## Description

3,366 definitions in this repository carried no docstring, across 323 of 391 Python files. That is
what stops a docstring check being enforced over the whole tree rather than only on changed code.
Every one of them now has a docstring, written from the code it describes.

The change is docstrings only. Nothing else moved, and that is checked rather than asserted: strip
every docstring from the tree and the syntax tree is identical to the parent commit, and the diff
carries **zero deleted lines**.

### Related Tasks

- Task ID: jHccIimO
- Follows on to the docstring check on task FmnlFpsh, which can move from diff-scoped to repo-wide
  once this lands

### Type of Change

- [x] Chore / maintenance

### What does this PR change?

| Area | Docstrings added | Notes |
|---|---|---|
| Shipped source (`swarm/`, `miner/src`, `neurons/`) | **1,628** | modules, classes, methods, nested helpers, private functions |
| Test suites (`validator/tests`, `miner/tests`) | **1,727** | each states what the test proves, not its steps |
| **Total** | **3,355** | across 322 files |

- **Coverage is the rule, not the public API.** Nested closures, `_private` helpers and dunder
  methods are included, since those are the ones a reader arrives at with no context.
- **Test docstrings state the guarantee.** `A symlink hashes its target, never the bytes it
  resolves to.` rather than `Tests the hash function.`
- **`swarm/submission_manifest/__init__.py` is untouched**, its 11 definitions deliberately left
  without docstrings. `test_backend_mirror_matches_apart_from_imports` asserts that file stays
  text-identical to `swarm-backend/app/submission_manifest.py` apart from import roots, so a
  docstring there has to land in both repositories in the same change.
- **No existing docstring was edited.** Not reworded, not re-wrapped, not repunctuated.

### Run it yourself

```bash
```

To confirm the claim that nothing but docstrings changed, without running anything:

```bash
git diff --numstat origin/main...HEAD | awk '$2 != 0 {print "deleted lines in " $3; bad=1}
  END {print bad ? "not docstrings only" : "insertions only, across " NR " files"}'
```

### How was this tested?

- `pytest validator/tests miner/tests -n4` in the CI image, on `origin/main` first and then on this
  branch: **1179 passed, 92 skipped** both times, identical.
- `pre-commit run --all-files`: **all 14 hooks pass**. Three misspellings the new text introduced
  were caught by codespell and fixed.
- `python3.11 -m compileall`: **391 of 391** files compile.
- Docstring-only proof, per file: parse the before and after, strip every docstring from both,
  compare the syntax trees. **322 of 322 identical.** A second check confirms every original line
  is still present byte for byte, which the syntax-tree check cannot see: it caught eight cases
  where inserting a docstring clipped a space out of the line below it.

<details>
<summary>Accuracy review of all 3,355 docstrings</summary>

Presence is mechanical; truth is not. Every added docstring was then read back against the code it
describes by a reader who had not written it, and **120 were corrected**:

```
 44  wrong         states something the code does not do
 40  overclaims    promises more than the code delivers
 35  misleading    defensible, but leaves the wrong mental model
  1  vague
120  total, 3.6% of the added docstrings
```

Representative corrections:

```
victim spawn point described as the drone's spawn point
terrain slope documented as degrees from vertical; it is degrees from horizontal
retry backoff documented as capping at 300s; min(300, 2**min(n,8)) caps at 256
loading-dock shutter gap stated backwards: closed adds 2.8m, open adds 0.0
overview camera said to orbit the goal; it orbits the drone-goal midpoint
a test claiming bit-exact determinism where the assertion is math.isclose at 1e-12
```

</details>

No. Nothing executes differently. A miner reading the submission template or a contributor reading
the validator now finds a line of prose on every definition.

### Risk & Rollback

The risk is a docstring that reads authoritatively and is wrong, which is why the accuracy pass
above exists and why its corrections are counted rather than summarised. Runtime risk is nil: no
statement changed, and any module reading its own `__doc__` would have shown up in the suite.
Revert the single commit to undo it.

### Documentation

- [x] The code is the documentation here, and this is the change.


robojules (approved):
LGTM

PR #148, opened 2026-09-13
Remove the causes of the test warnings instead of filtering them

S3333F:
## Description

A full test run printed 96 warnings. Two causes accounted for all but five of them, and both were
real: one meant 21 tests had no timeout despite asking for one, the other meant a background
thread crashed into whichever test happened to be running fifteen seconds later.

Neither is filtered. The causes are gone.

### Related Tasks

- Task ID: cS9ptAjg
  https://github.com/swarm-subnet/swarm-gym-pybullet-drones/pull/3, which removes the last
  warning. Independent of this one and not required by it.

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| `requirements.txt` | `pytest-timeout==2.4.0` | 21 tests now actually time out |
| `test_streaming_phase.py` | 8 tests stop the heartbeat timer they start | Nobody, the suite stops lying |
| `validator/tests/conftest.py` | A fixture that fails any test leaving a timer alive | Whoever writes the next one |
| `heartbeat.py` | `on_seed_complete` guarded like `_timer_loop`; the timer thread is named | Nobody |
| `.github/workflows/tests.yml` | Rebuild the image when `requirements.txt` changes, not only the Dockerfile | Nobody |

**The unknown marker was not cosmetic.** 21 tests carry `@pytest.mark.timeout(...)`, pytest had no
implementation for it, so each warned and then ran with no limit at all. A hanging test would have
sat there until the job's own 30 minute timeout killed the whole run.

**The leaked thread explains a failure that looked random.** Eight tests call `hb.start()` inside
`asyncio.run()` and never stop it. `asyncio.run` closes the loop when the test ends; the timer
thread wakes 15 s later and posts to the closed loop, so `RuntimeError: Event loop is closed` is
reported against an unrelated test, and under `-n4 --dist worksteal` a different one each run.
They stop the timer in a `finally` now. The thread is named, and an autouse fixture fails any test
that leaves one running, so the next leak points at itself instead of at a bystander.

`on_seed_complete` makes the same `call_soon_threadsafe` call from a worker thread and now carries
the same guard as the timer loop.

**The image rebuild was a trap waiting to happen:** a dependency added to `requirements.txt` was
absent from the image the tests ran in, because the rebuild only triggered on `.docker/Dockerfile`.
This PR is the first change that needed it. The run on this commit rebuilt in 1m43s rather than
skipping in 9s, so the trigger works.

### One warning is left on purpose

```
swarm/__init__.py:35: DeprecationWarning: pkg_resources is deprecated as an API
```

`pkg_resources` is imported for the drone gym, which still depends on it. Removing it means
releasing a new version of `swarm-drone-gym`, a separate package, which is what the linked PR
does. That is deliberately not bundled here. Five of the 96 are this one line.

### Run it yourself

```bash
```

To watch the guard work, delete the `finally` from any of the eight tests and run that file: it
fails by name instead of poisoning a different test later.

### How was this tested?

- `pytest miner/tests` and `pytest validator/tests` in `ghcr.io/swarm-subnet/swarm:base`:
  **1,180 passed, 92 skipped, 0 failed.**
- Warning count over the same run, before and after:

| | Before | After |
|---|---|---|
| `PytestUnknownMarkWarning` | 21 | **0** |
| `PytestUnhandledThreadExceptionWarning` | 16 | **0** |
| `DeprecationWarning` (drone gym) | 1 | 1 |
| **Total occurrences** | **96** | **5** |

- `pre-commit run --all-files`: all hooks pass.

No. Nothing at runtime changes. `on_seed_complete` returns quietly instead of raising if the loop
has closed under it, which in a real validator cannot happen: both callers stop the heartbeat in a
`finally` before the loop goes.

### Risk & Rollback

The new fixture is the only thing that can fail a previously passing test, and only by catching a
leak that was already breaking runs at random. Revert the commit to undo all of it.

### Documentation

- [x] Not applicable: no document describes the test suite's warning policy.


robojules (changes requested):
I would prefer another fix rather than  ignore the deprecation warning, I would recommend:

Step 1: fix the gym (swarm-subnet/swarm-gym-pybullet-drones)

Make the same change in all three files:


- import pkg_resources
+ from importlib.resources import files

# gym_pybullet_drones/envs/BaseAviary.py  (lines 486 and 989)
- pkg_resources.resource_filename('gym_pybullet_drones', 'assets/'+self.URDF)
+ str(files('gym_pybullet_drones') / 'assets' / self.URDF)

# gym_pybullet_drones/control/BaseControl.py:202 and CTBRControl.py:236
- path = pkg_resources.resource_filename('gym_pybullet_drones', 'assets/'+URDF)
+ path = str(files('gym_pybullet_drones') / 'assets' / URDF)
str() gives a real file path because pip installs wheels unpacked, and that's what p.loadURDF and etxml.parse need. Then bump the version to 2.0.0.4 in the gym's pyproject.toml and publish it.

Step 2: clean up this repo

Change the pin in [requirements.txt](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/requirements.txt) to swarm-drone-gym==2.0.0.4.
In [swarm/init.py](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/__init__.py#L33-L48), delete the whole block: the warnings import, the catch_warnings wrapper, and the pkg_resources shim. Nothing imports pkg_resources any more, so there's nothing to hide or fake.
Keep [validator/tests/test_runtime_compat.py](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_runtime_compat.py). It checks that the gym imports when pkg_resources is missing. Once the shim is gone, it passes only if the gym really doesn't depend on pkg_resources, so it guards against the old import coming back.

robojules:
Also, I would like to fix this:

```
 Traceback (most recent call last):
    File "/usr/local/lib/python3.11/threading.py", line 1045, in _bootstrap_inner
      self.run()
    File "/usr/local/lib/python3.11/threading.py", line 982, in run
      self._target(*self._args, **self._kwargs)
    File "/app/swarm/validator/utils_parts/heartbeat.py", line 136, in _timer_loop
      self.main_loop.call_soon_threadsafe(
    File "/usr/local/lib/python3.11/asyncio/base_events.py", line 807, in call_soon_threadsafe
      self._check_closed()
    File "/usr/local/lib/python3.11/asyncio/base_events.py", line 520, in _check_closed
      raise RuntimeError('Event loop is closed')
  RuntimeError: Event loop is closed
  
```

Why this error appears:

s: 8 tests in [test_streaming_phase.py](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1222) start a heartbeat and never stop it. test_swarm_sar_rollout_scores_for_random_count isn't the cause; it was just the test running when the error surfaced.

Each of these tests calls hb.start() inside asyncio.run(_run()) and never calls hb.finish(). start() launches a background timer thread that sends a heartbeat every 15 s ([heartbeat.py:123-126](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/validator/utils_parts/heartbeat.py#L123-L126)).
asyncio.run closes the event loop when the test ends, but the timer thread keeps running.
About 15 s later the thread wakes up and tries to schedule a heartbeat on the closed loop. That raises RuntimeError: Event loop is closed.
pytest reports the error against whichever test is running at that moment. With -n4 --dist worksteal in CI, that's a different test on each run, which is why it looks unrelated.
I confirmed this by running test_run_task.py, test_streaming_phase.py and test_heartbeat_and_fallback.py in the local ghcr.io/swarm-subnet/swarm:base image with a small leak-detection plugin (only in the scratchpad, not in the repo). All 66 tests passed. Exactly these 8 tests left a timer thread running:

test_heartbeat_manager_honors_stop_required ([line 1222](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1222))
test_heartbeat_manager_start_resets_stop_flag ([line 1261](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1261))
test_heartbeat_manager_ignores_none_response ([line 1439](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1439))
test_heartbeat_manager_ignores_response_without_stop_required ([line 1453](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1453))
test_heartbeat_manager_stop_without_conflicts_uses_default_reason ([line 1467](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1467))
test_heartbeat_manager_handles_post_exception ([line 1481](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1481))
test_heartbeat_manager_ignores_stale_session_response ([line 1495](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1495))
test_heartbeat_manager_stop_latches_until_next_session ([line 1512](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/validator/tests/test_streaming_phase.py#L1512))
The production code is fine. Both places that start a heartbeat stop it again in a finally ([evaluation.py:712-713](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/validator/utils_parts/evaluation.py#L712-L713) and [evaluation.py:900-901](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/validator/utils_parts/evaluation.py#L900-L901)), so in a real validator the loop never closes while the timer is still running.

How to fix it: make the tests stop what they start. In each of the 8 tests, wrap the body in try/finally and stop the timer:


hb = HeartbeatManager(_Api(), asyncio.get_event_loop())
hb.start("evaluating_screening", uid=66, total=200)
try:
    await hb._safe_heartbeat(0, hb._session_id)
    assert hb.should_stop() is None
    await hb._safe_heartbeat(10, hb._session_id)
    return hb.should_stop()
finally:
    hb._stop_timer()
Use **_stop_timer()** rather than **finish(). finish()** also sends a final heartbeat and an "idle" heartbeat through the fake API, which some of these tests count or read responses from. In the tests that call start() twice, one finally at the end is enough, because start() already stops the previous timer.

What this means for PR #148: the try/except RuntimeError it adds in _timer_loop only hides the error. The threads still leak, and its new test covers a situation that only these leaking tests create. I'd fix the 8 tests instead, and then keep the except as extra protection or drop it. If you keep it, note that on_seed_complete ([heartbeat.py:165](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/validator/utils_parts/heartbeat.py#L165)) makes the same unprotected call_soon_threadsafe call.

To stop this coming back, an autouse fixture in conftest.py could fail any test that leaves a _timer_loop thread running. That points at the leaking test itself, not whichever test happens to be running 15 s later.

robojules:
PendingDeprecationWarning: Please use `import python_multipart` instead.
    import multipart 


Implementa lso this :arrow_double_up:

robojules (approved):
LGTM, amazing job! :rocket:

PR #149, opened 2026-09-13
Check docstrings on the code a change touches

S3333F:
## Description

Nothing in the repository checks that a new function carries a docstring, so the standard holds
only when someone remembers it in review.

The check now runs in pre-commit and in the Lint job, and it reads the diff rather than the tree.
3,355 definitions across 323 files still have none, so a whole-tree check would fail every pull
request over code nobody in it touched. A definition has to carry a docstring once the branch
adds or changes one of its own lines.

### Related Tasks

- Task ID: FmnlFpsh
- Follows on to #147, which writes the missing docstrings. Once that lands, this moves from the
  diff to the whole tree by adding `--all` to the hook.

### Type of Change

### What does this PR change?

|---|---|---|
| `scripts/check_docstrings.py` | New. Diff-scoped by default, `--all` for the whole tree | Contributors who forget one |
| `.pre-commit-config.yaml` | A local hook that takes no file list, so CI running `--all-files` cannot widen its scope | Contributors: one more hook |
| `.github/workflows/lint.yml` | `fetch-depth: 0`, since a shallow clone holds no merge base to diff against | Nobody |
| `validator/tests/` | 12 tests for the check, plus one line in the `parents[n]` path registry | Nobody |

- **Every definition counts, not only the public ones.** Underscore-named functions, functions
  written inside another function, methods, classes and modules. ruff's `D1xx` rules see neither
  of the first two, which is why this is a script rather than three lines of config.
- **A changed line is attributed to the narrowest definition holding it.** Editing one line
  inside a nested helper asks for a docstring on that helper, not on every function around it.
- **Silent where nothing was written.** Deleted lines, blank lines, whole-line comments and pure
  renames put nothing in scope.
- `swarm/submission_manifest/__init__.py` is excluded. A test pins it text-identical to the
  backend copy apart from import roots, so a docstring there has to land in two repositories in
  the same change.

### Run it yourself

Add an undocumented helper and watch the hook name it:

```bash
printf '\n\ndef _tally(rows):\n    return len(rows)\n' >> swarm/utils/hash.py
git add swarm/utils/hash.py
pre-commit run docstrings
# swarm/utils/hash.py:53: no docstring on _tally
git restore --staged swarm/utils/hash.py && git checkout -- swarm/utils/hash.py
```

And the suite:

```bash
```

### How was this tested?

- `pytest miner/tests` and `pytest validator/tests` in `ghcr.io/swarm-subnet/swarm:base`:
  **1,192 passed, 92 skipped**, against **1,179 passed, 92 skipped** on the parent commit. The 13
  new cases are the 12 tests for the check and one parametrised entry in the path registry.
- `pre-commit run --all-files`: **all 15 hooks pass**, the new one included.
- Scope measured from both ends: `--all` reports **3,355** on this branch and **0** against the
  branch behind #147. The flip to whole-tree is therefore green the day that merges, with nothing
  left over to write.
- Behaviour on real files: changing one line inside a nested helper in
  `swarm/benchmark/engine_parts/workers.py` reports that helper alone, and neither the function
  wrapped around it nor the module. A `git mv` of the same file reports nothing at all.

No. Nothing at runtime changes. A contributor who adds a definition without a docstring now sees
the Lint job fail with the file, the line and the name.

### Risk & Rollback

The risk is stopping a change that wrote nothing, which is why deletions, blank lines, comments
and renames are all out of scope. Nothing in the product imports the script. Revert the commit,
or drop the hook block to switch it off while keeping the script.

### Documentation

- [x] Not applicable: no document in the repository describes the code style, and the rule is
  stated in the check's own failure message.


robojules (approved):
LGTM

PR #151, opened 2026-09-14
Physics mode per family, ground effect height from the altitude ray

S3333F:
## Description

The drone model carries air drag, ground effect and downwash with coefficients in its URDF, but every family runs the plain physics mode where none of them apply. Families can now opt into the aerodynamic terms with one class attribute; every existing family stays on the plain mode and its flights and images are unchanged. The ground effect in the drone library measures height as world z, which is wrong on any map whose floor is not at z = 0; the env now reads the height from the altitude ray it already casts.

### Related Tasks

- Task ID: uBvBPpmB

### Type of Change

- [x] New feature

### What does this PR change?

|---|---|---|
| Family runtime | `physics_mode` class attribute, default **"pyb"**, beside `seeded_sun` and `render_backend` | Nobody today; a new family sets it |
| Env | `physics=None` resolves to the family's mode; an explicit argument still wins | Nobody |
| Ground effect | Height from the downward altitude ray instead of world z, scalar math | Opt-in families only |
| Wind | With wind on, the library's still-air drag is skipped: the wind force already applies the same rotor drag on the relative air | Opt-in families only |

- The coefficients stay in each drone's URDF (`gnd_eff_coeff`, `drag_coeff_xy`, `drag_coeff_z`, `dw_coeff_1..3`, `prop_radius`); nothing drone-specific is in code.
- The office family keeps its own hand-written drag and ground cushion and stays on the plain mode, so nothing stacks.
- Five new tests: every family stays plain, the env takes the family's mode, drag shortens a full-speed flight, the cushion follows the ray height, wind replaces the still-air drag term.

### Run it yourself

```bash
```

The new tests alone:

```bash
pytest validator/tests/test_family_physics_mode.py -n 0
```

### How was this tested?

All numbers measured on the same 12-core box with the master wheel (2.0.0.5), fresh venv, 50 Hz, one drone on the open map.

- `pytest validator/tests -n4 --dist worksteal`: **1151 passed, 77 skipped**. `pytest miner/tests -n 0`: **47 passed**.
- Existing families unchanged: fixed action stream, 150 steps, one seed per family on main and on this branch, per-step positions hashed: **7 of 7 families identical**. `verify_render_identity.py`: **9 of 9 hashes identical**.
- Air terms on, straight flight at 1.5 m/s for 10 s:

| | Plain | Air terms on | Difference |
|---|---|---|---|
| Distance flown | **14.63 m** | **14.25 m** | 0.39 m (2.6 %) |
| Speed at 10 s | 1.497 m/s | 1.468 m/s | 2 % slower |
| Pad landing | success, **9.30 s** | success, **9.54 s** | 0.24 s later |
| Touchdown vertical speed | 0.17 m/s | 0.14 m/s | softer |
| Force calls per substep | 17 us (thrust) | 17 + 53 us | +0.05 ms |
| Whole control step, camera included | 5.06 ms | 5.28 ms | within round noise |

- Ground effect height: on three mountain seeds the pad sits at world z of **17 to 46 m**, so the library formula gives **0 %** cushion on the pad; the ray reads **0.07 m** and gives **7.7 %**, the same as on the flat open map.

<details>
<summary>Step cost, five alternating rounds of 1000 steps</summary>

```
plain  5.06  4.34  4.35  5.39  5.89  ms   median 5.06
aero   5.03  4.45  5.28  5.38  5.82  ms   median 5.28
```

</details>

Nobody yet. A family that sets `physics_mode = "pyb_gnd_drag_dw"` flies about 2 to 3 % shorter at cruise and gets a small cushion within a few prop radii of a surface, so its scores differ from the plain mode. Existing families keep their exact flights.

### Risk & Rollback

The default path is untouched, proven by the trajectory and render hashes. Revert the commit, or leave the attribute at its default.

### Documentation

- [x] The attribute is documented where it is declared, next to the two sibling switches; no separate doc needed.


robojules (approved):
LGTM

PR #153, opened 2026-09-14
Let a family render the sky of its seeded sun

S3333F:
## Description

A family that opted into the seeded sun still renders a white background, and the only alternative is a fixed two-colour gradient that never follows the sun. Two new family switches let the renderer compute the sky from the seed's sun instead, with seeded clouds if wanted. Nothing is on for any family: every registered family keeps its exact pixels, and a test says so.

### Related Tasks

- Task ID: Lw1lUuIR
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/bullet3-swarmfork/pull/22 (the renderer side; needs that wheel, merge as a pair)

### Type of Change

- [x] New feature

### What does this PR change?

- **Family runtime**: `sky_from_sun` and `sky_clouds`, both default `False`, next to `seeded_sun`.
- **Env**: `_apply_sun_sky` runs once per episode after the sun is chosen. With `sky_from_sun` and a daytime sun it sets `ER_SWARM_SKY_SUN` on the three colour camera calls (observation, on-demand RGB, video), and with `sky_clouds` it passes the map seed as `skyCloudSeed`. A sun that says it is night (`night` on the sun, arriving with the day-and-night card) keeps the family's own sky. An old wheel without the flag raises instead of rendering white.
- **Tests**: `test_sky_colors.py` covers the switches and the kwargs; `test_sky_sun.py` pins the rasterised and ray-cast sun-sky frames of the test world to committed SHA-256 hashes and reruns them at 1, 2 and 4 render threads.

### Run it yourself

```bash
```

### How was this tested?

- `pytest -q -m "not slow and not integration"` on the branch wheel: **1217 passed, 76 skipped**.
- `validator/scripts/verify_render_identity.py`, 7 family configs, wheel of current master against the branch wheel: **identical hashes on every config**, rasterised path at 2 threads, ray-cast path at 1, 2 and 4 threads.
- Sun-sky frame hashes at 1, 2 and 4 threads: **one value per path**, now committed in `test_sky_sun.py`.

### Risk & Rollback

No family opts in, so no observation changes; the new code runs only when a family sets `sky_from_sun`. Revert the commit.

### Documentation

- [x] The two switches are documented where they live, next to `seeded_sun`; the family guide does not list runtime switches.


robojules (commented):
As soon as the tests passes, LGTM!

robojules (approved):

PR #154, opened 2026-09-15
Pin the sun sky frames to the sun the seed draws today and check the blue sky under a high sun

S3333F:
## Description

`test_sky_sun.py` pins the sun-sky frame of the test world to two hashes and checks that the sky it paints is blue. Both were written under the sun that seed 101 drew at the time: 13:20, 55 degrees high, white. The day-and-night change then redrew how a seed picks its sun, and seed 101 now gives 16:34, 23 degrees, orange. Under that sun the frames hash differently and the sky is a sunset glow, so the four tests fail on any wheel that carries the sun sky. Nothing saw it: the wheel those tests need is not the pinned one, so they skip in CI, and the day-and-night change was checked against a wheel without the flag.

The hashes are now the ones the current sun produces, and the blue-sky check renders its own frame under a seed whose sun stands high, with the elevation asserted so the next change to the sun distribution fails with a clear message instead of a colour comparison.

Fixes # (issue)

### Related Tasks

- Task ID: Lw1lUuIR

### Type of Change

- [x] Bug fix
- [x] Test coverage

### What does this PR change?

- `validator/tests/test_sky_sun.py`: the two committed SHA-256 values, a `HIGH_SUN_SEED` for the blue-sky check with an elevation assertion, and a `seed` argument on the frame helper. Tests only; no runtime code changes.

### How was this tested?

- Commands run, on a 12-core AMD EPYC box, swarm-bullet3 built from the fork's current `master` (`2438ded`), `pytest -q validator/tests/test_sky_sun.py validator/tests/test_sky_colors.py validator/tests/test_daylight.py` at 1, 2 and 4 render threads; `pre-commit run --files validator/tests/test_sky_sun.py`.
- Result: **30 passed** (4 failed on `main` with the same wheel); the sun-sky frame hashes are the same at 1, 2 and 4 threads on both paths; pre-commit clean. On the pinned 2.0.0.3 wheel the file skips as before.

| Seed 101 sun | Before the day-and-night change | Now |
|---|---|---|
| Hour, elevation | 13:20, 55.1 degrees | 16:34, 22.8 degrees |
| Colour | white | (1.00, 0.79, 0.60) |

None. Tests only.

### Risk & Rollback

None at runtime. If the pinned hashes were wrong the test would fail on the next wheel, which is what it is for. Revert the commit.

### Documentation

- [x] Not applicable (explain why): test-only change, the constants are commented in the test.


robojules (approved):
LGTM, good catch!

PR #155, opened 2026-09-15
Document the family options and switches that steer the simulation engine

S3333F:
## Description

Seven family options, a per-map wind table, a collision cache flag and six environment variables now steer the simulation engine from this repository, and every one of them is off, so every registered family renders and flies as it did before they existed. They are documented where they are declared and nowhere else, so a person or an agent starting a new map has to read the environment to learn what can be switched on and which engine flag each option turns into. `docs/simulation.md` lists them with their defaults, the engine switch each one uses, the four render calls they land in, the seed-to-light-to-sky chain, the environment variables and constants, the steps a new map follows to get the full picture, and the tools and tests that prove nothing moved.

Fixes # (issue)

### Related Tasks

- Task ID: E0znd931
- Related tasks/PRs (other repos): the engine side of the same reference, `docs/swarm_api.md`, is in https://github.com/swarm-subnet/bullet3-swarmfork/pull/25 (independent; each document points at the other)

### Type of Change

- [x] Documentation

### What does this PR change?

- `docs/simulation.md`: new. Family options, per-map options, environment variables, constants, the render calls, the seed to light to sky chain, the new-map steps, tools and tests. Names and defaults come from `swarm/challenge_families/base.py`, `swarm/core/moving_drone.py`, `swarm/core/daylight.py`, `swarm/core/wind.py`, `swarm/constants.py` and `swarm/validator/seed_manager.py` on this branch; costs come from the measurements in the pull requests that added each option.
- `README.md`: one line under The Worlds pointing at the page.

### How was this tested?

- Commands run: a script that parses the `ChallengeFamilyRuntime` fields and the engine-related environment reads and checks each has a row (all present); an independent read of the document against the source for wrong claims; `pre-commit run --files docs/simulation.md README.md` (markdownlint included).
- Result: every field and variable has a row, the findings of the independent read are folded in, pre-commit clean.

### Tools & Dependencies

- Tools updated: none

None. Documentation only.

### Risk & Rollback

Low risk: no code changes. Revert the commit.

### Documentation

- [x] README.md updated
- [x] `docs/` updated


robojules (approved):
LGTM 🚀

PR #156, opened 2026-09-15
Guard test that keeps the simulation reference complete against the source

S3333F:
## Description

`docs/simulation.md` lists the family options and the engine-related environment variables that steer the simulation. A list kept by hand falls behind the first time a pull request adds an option and forgets the row. A test now reads the annotated attributes of `ChallengeFamilyRuntime` and the `SWARM_` reads in the simulator package, the seed manager and the worker module straight from the source, and fails with the name of anything that has no row. Names come from parsing, never from a list in the test.

Writing it caught one variable the document had missed, `SWARM_FOREST_FILE_VISUALS_ONLY`, which has its row now.

Fixes # (issue)

### Related Tasks

- Task ID: JQzuAADz
- Related tasks/PRs (other repos): the fork side, https://github.com/swarm-subnet/bullet3-swarmfork/pull/26. Both build on the reference documents of #155 and https://github.com/swarm-subnet/bullet3-swarmfork/pull/25, which are the base branches until those merge.

### Type of Change

- [x] Test coverage
- [x] Documentation

### What does this PR change?

- `validator/tests/test_simulation_doc.py`: two tests, text only, any wheel. The scope of the environment scan is the simulator package plus the two validator modules that hand the engine its cache folder and its container start, stated in the test.
- `validator/tests/test_repo_root_paths.py`: the new file registered in the resolver table.
- `docs/simulation.md`: the missing row.

### How was this tested?

- Commands run: `pytest validator/tests/test_simulation_doc.py validator/tests/test_repo_root_paths.py`; the same after deleting the `SWARM_RENDER_BACKEND` row and after removing every `night_share` mention from the document; `pre-commit run --files` on the three files.
- Result: 32 passed; with the row deleted the environment test fails naming `SWARM_RENDER_BACKEND`, with the mentions removed the family test fails naming `night_share`; restored, all pass; pre-commit clean.

### Tools & Dependencies

- Tools updated: none

None.

### Risk & Rollback

Low risk: two tests and one document row. Revert the commit.

### Documentation

- [ ] README.md updated
- [x] `docs/` updated


robojules (approved):
LGTM

PR #158, opened 2026-09-16
Derive epoch seeds from a backend key so every validator flies the same maps

S3333F:
## Description

Each validator rolls its own 1,100 seeds per epoch with `random.SystemRandom()`, and the backend
leases only seed *indexes*, so index N is a different map on every validator. Two models in one
epoch are never measured on the same maps, and re-running one model gives a different answer.

The seed list now comes from a key the backend holds, so every validator derives the identical
list and index N is the same mission everywhere.

### Related Tasks

- Task ID: xKdSFcca
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/swarm-backend/pull/39 (deploy
  that first; this PR is inert until a backend serves keys)

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Seed source | Derived from the epoch key instead of `SystemRandom` | Miners: everyone flies the same maps |
| Readiness | No key means no seeds and no work, never a fall back to random | Operators: a keyless validator idles instead of diverging |
| Stored seeds | Derived files stamped with the key's fingerprint, rejected on mismatch | Nobody: closes a silent wrong-maps path |
| Screening epoch | Screening reads the task's epoch instead of today's | Nobody: a bug on the rollover path |
| Updater | Runs from a copy, and a failed update no longer kills the watcher | Operators: a failed update retries instead of stranding the host |
| Seed count | **1,100 -> 1,000** an epoch; screening keeps its 300, so the benchmark portion is 700 | Miners: a shorter epoch benchmark |
| Version | `5.1.5.6` -> **`5.1.5.7`** | Operators: the fleet installs this, the backend sees no change |

**The version is the fourth digit on purpose.** `BENCHMARK_VERSION` is the first three
components, so this deploys to every validator while the backend's fence, its submission filter
and its version transition all see `5.1.5` and do nothing. Scoring carries on unchanged. The
switch to the shared scheme is a separate one-line release to `5.1.6.0`, after the fleet is
confirmed to be running this.

- **The derivation is pinned** and matches the backend byte for byte:
  `seed = HMAC-SHA256(key, "v1|<family>|<epoch>|<index>")[:4]`, key hex-decoded to 32 raw bytes.
  Committed known-answer vectors fail loudly if the encoding ever moves, instead of silently
  changing every map in the subnet.
- **A task carries the key for its own epoch**, adopted before the model is downloaded. A long
  poll that crosses a rollover returns a task for the epoch it was queued in, and it now flies
  that epoch's list.
- **`state/` survives an update**, so a validator that upgrades mid-epoch still has the seeds it
  rolled before the switch sitting on disk, and the old loader accepted them: same epoch, same
  family, same count. Every derived file now carries the key's fingerprint and is discarded when
  it does not match. Unstamped files from the random scheme are deliberately kept, because no key
  can rebuild them.
- **The key is checked against its published commitment** before it is used.
- **The epoch drops to 1,000 seeds.** The map mix barely moves: every family stays within 0.1 of
  a percentage point of its old distribution, and the 100-slot benchmark templates still tile 700
  exactly. `swarm/templates/README.md` is deliberately left saying 1,100: the backend pins its
  hash and admission requires the current one, so changing it would reject every miner who has
  not re-copied the template.
- **The validator logs the seed-set identity it built.** The fleet can be checked without
  switching anything on: read that line off every validator and the strings either all match or
  they do not. It is a hash of the commitment, so it names the list without revealing a map.

> [!IMPORTANT]
> `update_deploy.sh` runs `git reset --hard`, which overwrites that script while bash is still
> reading it. Bash continues from a byte offset that now points at different text. Reproduced
> both outcomes: executing garbage and exiting **127**, and stopping partway while exiting
> **0** and reporting success. In the second case the install and the restart never happen, the
> checkout already reports the new version, so the watcher sees nothing left to do and that host
> stays on old code with every log line claiming the update worked. The script now re-executes
> from a temporary copy, so the running bytes can no longer move.

### Run it yourself

```bash
```

### How was this tested?

- `pytest validator/tests`: **1203 passed, 96 skipped**, at 1,000 seeds. Two results are not green and both
  reproduce on an untouched checkout of `origin/main` on the same host:
  `test_submission_runtime_caps` (a `pkg_resources` import in that venv) and `test_packaging`
  (the `wheel` package is absent there).
- **18 new tests** covering the derivation and the seed manager under the shared scheme.
- The derivation was compared against the backend implementation over 8 combinations of key,
  epoch and family across all 1,100 indexes: identical throughout.
- The self-overwrite failure and its fix were reproduced directly:

<details>
<summary>Script overwriting itself, before and after</summary>

```
without the copy:
  line 1 ran
  t2.sh: line 6: old: command not found
  EXIT: 127

  (a smaller replacement instead stops silently after line 1 and exits 0)

with the copy:
  line 1 ran
  line 2 ran
  line 3 ran (reached the end)
  EXIT: 0
```

</details>

This repository's suite can no longer run on a Python 3.10 host, because `swarm-worlds` requires
3.11; the existing tests fail there identically. The run above is from a 3.11 host.

Not yet tested end to end: no validator has derived a list from a key a real backend served. That
is the first step of the rollout, on our own validator, before anything is switched on.

Not on merge. The shared scheme stays off until the subnet runs benchmark version `5.1.6`, and
this release keeps the third digit at `5.1.5`. Validator operators get the repaired updater and
nothing else. When the scheme is switched on, miners are all measured on the same maps and can
rebuild an epoch's maps from the published key to check their own scores.

### Risk & Rollback

The derived path is unreachable while the subnet is on `5.1.5`, so the behaviour change on merge
is limited to the two updater scripts. The updater is also where the risk sits: this is the one
update performed by the old, self-overwriting script, which is why it ships on a version that
changes nothing about scoring. Every validator's running version is checked afterwards, and a
host that did not take it is re-run by hand. Rollback is a revert; validators only update upward,
so recovery from a later release is a roll-forward.

### Documentation

- [x] `validator/docs/validator.md` rewritten: the "Per-Validator Seeds" section described the
  old behaviour end to end
- [x] `README.md` and `miner/docs/miner.md` corrected, including how a miner verifies a score
  after an epoch closes


miguelik2 (approved):
LGTM

robojules (approved):
LGTM

PR #159, opened 2026-09-17
Ship the validator as a published image instead of a host install

S3333F:
## Description

An operator's machine currently decides what the validator runs on. They clone the repo,
install system packages, build a virtualenv and let the updater `git reset` and reinstall on
every release, so each host ends up with whatever its own package resolution produced that
day. Two validators running the same commit are not running the same software.

The validator now ships as a published image. Operators pull it; nothing is compiled or
installed on their side, and every validator runs an identical package set. The same image
runs any command without a wallet, which is what a test job or a nightly build needs from it.

### Related Tasks

- Task ID: sulazhTY

### Type of Change

### What does this PR change?

|---|---|---|
| Image | `.docker/validator.Dockerfile` on `ghcr.io/swarm-subnet/swarm:base`, with its own uv, the Docker CLI, buildx, and capabilities on `nsenter` and `iptables` | Nobody yet: nothing runs it until CI publishes |
| Requirements | `validator/requirements.txt`, the runtime set only; a test fails if any pin differs from the root file | Nobody: same versions, shorter list |
| Compose | A `validator` service behind a `validator` profile: `up` pulls the published image, `build` builds it from the validator Dockerfile on the current base | Operators: one command to run; anyone: a local build that works |
| Container start | Runs as a plain uid: `HOME` is the state directory, the wallet is a read-only mount at `/wallets`, the code's own `state/` folders and the log land on the host | Operators: state survives an image update |
| Command mode | A command after the image name runs as-is, no wallet and no backend | CI and nightly builds |
| Deploy | `update_deploy.sh` pulls the image and starts the container instead of resetting and reinstalling; it re-runs itself from a copy and is told where the checkout is | Operators: nothing is built on their host |
| Auto-update | `auto_update_deploy.sh` compares the running image's version label against a freshly pulled `:latest` | Operators: no action, see below |
| CI | `validator-image.yml` publishes `:<version>` and `:latest` after **Tests** passes on `main`; a change to the image's inputs without a bump republishes `:latest` only | Nobody |
| Pin check | `env_check` accepts a build label the pin does not name, so `torch 2.10.0+cpu` from the base is not drift | Nobody |
| Check | `validator/scripts/cross_machine_check.py` scores one model on one pinned image across several hosts and compares | Us, before we trust the image |

- **Existing operators do nothing.** The watcher already running on their host checks a
  version and then executes `update_deploy.sh` *from disk*. That script now stops the host
  process and starts the container, so a host moves onto containers on its next update with
  nobody restarting anything. It still syncs the checkout, which is what keeps that watcher's
  version check meaningful rather than firing on every cycle forever.
- **One hotkey, one validator.** The deploy stops the PM2 host process before starting the
  container, and only after the image is in hand, so a failed pull leaves the host running.
  Without that a box runs both and the backend fences one as a duplicate session.
- **Identical mount paths where the daemon is involved.** `/tmp`, `/dev/shm` and the state
  directory are mounted at the same absolute path on both sides, because the validator hands
  those paths to the host's Docker daemon when it starts evaluation containers. The wallet and
  the code's `state/` folders never reach the daemon, so they are mounted where the uid can
  use them: a wallet under `/root` on the host is unreadable at that path by a plain uid.
- **Not root.** The container runs as an unprivileged uid with `SYS_ADMIN` and `NET_ADMIN`;
  `nsenter` and `iptables` carry file capabilities rather than the whole process being root.
  A bare uid has no passwd entry, so Docker sets `HOME=/` and bittensor dies creating
  `~/.bittensor` on import; compose sets `HOME` to the state directory and a plain
  `docker run` gets a scratch home.
- **The version label is the contract.** CI refuses to overwrite a version that is already
  published, because pinning `SWARM_VALIDATOR_TAG` is how a rollback works and a tag has to
  keep meaning one exact image. A Dockerfile-only change therefore reaches running validators
  at the next version bump, not before.
- **The runner image's inputs are untouched.** The validator image has its own
  `validator.Dockerfile.dockerignore`; the root `.dockerignore` is left byte-identical
  because the runner image's cache key hashes it, and editing it would rebuild the runner on
  every validator for nothing. A test holds that line.

### Does one image actually give one score?

A shared image is worth building only if it removes the difference. It may not: the same
machine already reproduces a flight tick for tick, which points at the CPU rather than at the
installed packages, and an image does not change the CPU.

`validator/scripts/cross_machine_check.py` settles it. Every host flies the same 100 frozen
seeds, with the same model, inside the image **CI already publishes** from the same
Dockerfile each validator builds locally. `compare` then puts the per-seed scores side by
side.

```bash
docker login ghcr.io -u <user> --password-stdin   # read:packages

# on each host
python validator/scripts/cross_machine_check.py run --model sar_champion --out host_a.json

# anywhere, once the reports are together
python validator/scripts/cross_machine_check.py compare host_a.json host_b.json
```

- **The pin has to survive the evaluator.** It rebuilds the runner unless the image carries
  a `swarm.code_hash` label matching the checkout, and it writes that label during its own
  build, so a pulled image has none. The published image was checked: it carries no labels
  at all. The script stamps the label after pulling, otherwise the pin is silently replaced
  by a local build and the check measures the thing it exists to rule out.
- **The registry digest is the evidence**, not the local image id, because stamping the
  label rebuilds the config and gives each host a different id for identical content.
- Each record also carries the CPU model and flags and the numpy, pybullet and torch
  versions. If two hosts did not run the same image, `compare` says so and refuses to
  attribute the difference to anything. A per-seed gap above **0.001** is reported as
  meaningful, since the crown moves on about +0.008.

### Run it yourself

No hotkey needed for any of this.

```bash
# build the image on your machine and run the validator tests inside it
docker compose -f .docker/docker-compose.yml --profile validator build validator
docker run --rm ghcr.io/swarm-subnet/swarm-validator:latest pytest validator/tests -n4

# the tests that cover the image, the compose service and the scripts
pytest validator/tests/test_validator_image.py validator/tests/test_env_check.py -v
```

To start the container itself: a throwaway wallet (`btcli wallet new_coldkey` /
`new_hotkey`), its names and the backend URL in `.env`, then
`bash validator/scripts/update/update_deploy.sh`. The logs show it load the wallet, reach
the chain and stop at "not registered", which is the whole container path.

### How was this tested?

- `pytest validator/tests` inside `swarm:base`, as CI runs it: **1228 passed, 100 skipped**.
- The container, started through compose on a dev box as **uid 1000** with a throwaway
  wallet: the entrypoint loads the wallet from `/wallets`, the pin check is silent, the
  validator reaches finney and stops at "not registered" with exit 0. Everything it wrote
  (`.bittensor/miners`, `state/`, `swarm-state/`, `logfile.log`) landed in the state directory.
- `docker run --rm --user 1000:1000 <image> python -c "import swarm, bittensor"`: works
  with no wallet, no backend and no `.env`.
- The image built through `docker compose build` from a checkout whose local `swarm:base`
  predates uv, which is the build the review hit: passes, because the Dockerfile now brings
  its own uv and compose pulls the current base first.
- **14 tests on the image and scripts**: the pin parity between the two requirements files,
  the Dockerfile installing the validator list, the update script finding the checkout from
  its copy, the publish workflow watching files that exist, compose building from the
  validator Dockerfile while `up` keeps pulling, every write mapping to the state directory,
  and the entrypoint running a plain command without a wallet.

Not yet tested end to end: the image has never been built by CI or pulled by a validator,
because the GHCR package does not exist until this merges. Doing that on one of our own
validators is the first step after it does.

Validator operators stop installing and start pulling. Nobody has to run anything for the
switch: the next update carries them over. After it, `docker compose logs -f validator` is
where the logs are, and pinning `SWARM_VALIDATOR_TAG` in `.env` holds or rolls back a version.

### Risk & Rollback

Nothing changes until CI publishes an image and a version is bumped, so merging this alone
moves nobody. The risky moment is the first update that runs the new deploy script, because
it is performed by the old one; a host that fails to take it keeps running as it does today
and is visible in its reported version. Rollback is pinning `SWARM_VALIDATOR_TAG` to a
previous version, or reverting the commit, which restores the git-based deploy.

**Prerequisite before this can publish:** the new GHCR package needs write access for this
repository, the same issue PR #134 hit.

### Documentation

- [x] `validator/docs/validator.md` rewritten: install, run, build it yourself, command mode,
  logs, auto-update, migration from a host install, and rollback by pinning a tag
- [x] `.env.example` extended with the wallet, netuid, uid and image-tag variables


validator/docs/validator.md
```diff
@@ -150,30 +157,26 @@ btcli wallet overview --wallet.name my_cold --subtensor.network finney
...
-source validator_env/bin/activate
-
-pm2 start neurons/validator.py --name swarm_validator -- \
-  --netuid 124 \
-  --subtensor.network finney \
-  --wallet.name my_cold \
-  --wallet.hotkey my_validator \
-  --logging.debug
+cd swarm
+docker compose -f .docker/docker-compose.yml --profile validator up -d validator
```
robojules, line 164: This will not work as the validator image must be pulled before, there is two way to fix it, or make sure people builds the image locally using the `validator.Dockerfile` or make sure the validator pulls the image from our github package, I prefer second but really open with the final decision

.docker/docker-compose.yml
```diff
@@ -33,6 +33,48 @@ services:
...
 
+  # The validator itself, as a container. Behind the "validator" profile so a dev box
+  # running `docker compose up` still only gets the test services above.
+  #
+  #   docker compose --profile validator up -d validator
+  #
+  # It pulls a published image and never builds: an operator's machine contributes
+  # nothing to what runs, which is the whole point of shipping an image.
+  validator:
+    image: ${SWARM_VALIDATOR_IMAGE:-ghcr.io/swarm-subnet/swarm-validator}:${SWARM_VALIDATOR_TAG:-latest}
```
robojules, line 44: I would rather prefer to build using the  .validator.Dockerfile, as it will make it work locally and we can also use the image from github packages if we previously push it using workflows

.github/workflows/validator-image.yml
```diff
@@ -0,0 +1,76 @@
...
+  contents: read
+  packages: write
+
+env:
+  IMAGE: ghcr.io/${{ github.repository_owner }}/swarm-validator
+
+jobs:
+  publish:
+    # workflow_run fires whatever the outcome was, so the result is checked here.
+    if: ${{ github.event_name == 'workflow_dispatch' || github.event.workflow_run.conclusion == 'success' }}
```
robojules, line 26: Can we implement the same as we did with the base image, that if a change is being done in the validator.Dockerfile, this image is also being pushed in the github packages?

robojules (changes requested):
PTAL, I miss a kind of specific "requirement.txt" for the valis, which contains the dedicated version https://github.com/swarm-subnet/swarm/blob/main/requirements.txt, I would create one dedicated requirement.txt just for the validators inside the validator folder. Great work btw!

robojules (changes requested):
PTAL, when executing `docker compose -f .docker/docker-compose.yml --profile validator up -d validator`  the following error appears:

```
 => ERROR [6/8] RUN uv pip install --system --no-cache -r /opt/swarm-validator/validator/requirements.txt     && uv   0.2s
------                                                                                                                     
 > [6/8] RUN uv pip install --system --no-cache -r /opt/swarm-validator/validator/requirements.txt     && uv pip install --system --no-cache --no-deps -e /opt/swarm-validator:
[+] up 0/1/sh: 1: uv: not found
 ⠸ Image ghcr.io/swarm-subnet/swarm-validator:latest Building                                                          1.8s
validator.Dockerfile:56

--------------------

  55 |     # installed without resolving the root requirements a second time.

  56 | >>> RUN uv pip install --system --no-cache -r /opt/swarm-validator/validator/requirements.txt \

  57 | >>>     && uv pip install --system --no-cache --no-deps -e /opt/swarm-validator

  58 |     

--------------------

failed to solve: process "/bin/sh -c uv pip install --system --no-cache -r /opt/swarm-validator/validator/requirements.txt    && uv pip install --system --no-cache --no-deps -e /opt/swarm-validator" did not complete successfully: exit code: 127

```

robojules (changes requested):
PTAL There are still two separated problems:

```
Verified. Two separate problems, both from the container running as a numeric uid with no passwd entry:

1. HOME=/ inside the container. Docker sets that for user: 1000:1000 when the image has no matching passwd entry. bittensor/core/settings.py:10-23 does HOME_DIR = Path.home() and then WALLETS_DIR.mkdir() + MINERS_DIR.mkdir() at import time — so it tries to create /.bittensor and dies before any config is read. BT_WALLET_PATH can't help; it's only consulted later (settings.py:154).

2. The wallet mount is :ro, but the validator needs to write under it. After HOME is fixed, [config.py:64-66](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/swarm/utils/config.py#L64-L66) creates {logging_dir}/{wallet}/{hotkey}/netuid124/validator and writes state.npz and the events log there. logging_dir defaults to ~/.bittensor/miners, which is inside the read-only mount — and your host miners/ is empty, so nothing exists to reuse. That would be the next crash.
```

.docker/docker-compose.yml
```diff
@@ -33,6 +33,72 @@ services:
...
+  #
+  #   docker compose --profile validator up -d validator
+  #
+  # `up` pulls the published image and does not build: an operator's machine
+  # contributes nothing to what runs, which is the whole point of shipping an image.
+  # The build block is for working on the image itself:
+  #
+  #   docker compose --profile validator build validator
+  #   docker compose --profile validator up -d --pull never validator
+  validator:
```
robojules: which is the difference between validator service and swarm_validator? Shouldn't be the same?

.docker/docker-compose.yml
robojules: done

robojules (changes requested):
Still having issues when:

`docker compose -f .docker/docker-compose.yml build validator`

`docker compose -f .docker/docker-compose.yml up validator`


```
swarm_validator_node  | [ERR] /opt/swarm-validator-state is not writable by uid 1000
swarm_validator_node  |       on the host: sudo chown -R 1000:1000 /opt/swarm-validator-state
swarm_validator_node exited with code 2 (restarting)
```

It can be fixed using:

Created the state tree under your home:

```
/home/julia-swarm/swarm-validator-state/{state,swarm-state}   # owned 1000:1000
Added [.docker/.env](vscode-webview://02mgj2nq9veh140959n51k974pil1tlokqupg4tnl6r6vive8kce/.docker/.env) setting SWARM_STATE_DIR=/home/julia-swarm/swarm-validator-state.
docker compose config confirms all three mounts and working_dir now point there, and the wallet mount already resolved to /home/julia-swarm/.bittensor/wallets where your oberon wallet is. The old /opt/swarm-validator-state was empty apart from two empty subdirectories, so nothing needed migrating — you can leave it there, unused.

Start it with:


docker compose -f .docker/docker-compose.yml --profile validator up -d validator
I didn't start it myself — that opens a validator session on your hotkey, and the deploy script warns that a second session on the same hotkey gets fenced off by the backend, so that's your call to make.
```

And when the tests in CI should be executed you can run 


```
- name: Run the validator tests
  run: |
    docker run --rm --cpus=4 \
      -v "$PWD":/app -w /app \
      -e PYTHONPATH=/app -e HOME=/tmp \
      --user "$(id -u):$(id -g)" \
      "$TEST_IMAGE" \
      pytest -v validator/tests -n4 --dist worksteal -p no:cacheprovider
```


Let me know if you have any doubt or question

robojules (approved):
LGTM 🚀  Amazing job!

PR #161, opened 2026-09-20
Hand a timed-out seed back instead of flying it twice

S3333F:
## Description

In seed flow a seed that hit the wall-clock timeout was flown a second time on the same host. A frozen seed therefore held a worker for two full timeouts before any other validator could try it: **840 s x 2 = 1,680 s** on autopilot. Production history matches to the second: the median hold of a seed returned without a result is **1,703 s**, and 77 of those 86 cases are autopilot.

In seed flow the pool is the retry, so a timed-out seed now goes back at once and the next validator takes it.

This targets `v5.1.6.1`, the branch collecting the next update.

### Related Tasks

- Task ID: 0K2v3j6d
- Related tasks/PRs (other repos): https://github.com/swarm-subnet/swarm-backend/pull/44 (the hand-out ceiling; it goes first, and this PR reads the numbers it adds to the claim reply)

### Type of Change

- [x] Bug fix

### What does this PR change?

- **Timeout retry**: `batch_timeout`, `batch_timeout_partial` and `seed_cancelled` no longer trigger the in-place retry when a seed feeder is active. The seed is marked `INFRA` as before, so it is never uploaded as a miner score, and the heartbeat hands it back.
- **Unchanged**: the quick retry for `container_start_failed` and `INFRA_DOCKER` stays, since it costs seconds rather than a full timeout. The RPC transport retry stays. Batch flow, which has no pool to fall back on, keeps its one timeout retry.
- **Claim loop**: when the backend says a granted seed has been handed out before, the validator logs `seed 40 is on hand-out 4 of 6 · 2 left before it closes unscorable` and records a `seed_rehanded` telemetry event. A backend that does not send the numbers changes nothing.

What one frozen seed costs a validator, from each family's own timeout (`600 s + per-seed budget`, one seed per batch):

| Family | Timeout | Held before | Held now |
|---|---|---|---|
| Autopilot | 840 s | **28 min** | **14 min** |
| Office Interceptor | 900 s | 30 min | 15 min |
| Search and rescue | 960 s | 32 min | 16 min |
| Swarm autopilot | 1,500 s | 50 min | 25 min |
| Swarm SAR | 2,100 s | 70 min | 35 min |

### Run it yourself

```bash
docker run --rm -v "$PWD":/app -w /app -e PYTHONPATH=/app -e HOME=/tmp \
  ghcr.io/swarm-subnet/swarm:base \
  pytest validator/tests/test_docker_evaluator.py validator/tests/test_seed_feeder_wiring.py \
  -n4 --run-full -p no:cacheprovider
```

`--run-full` matters: the retry tests in `test_docker_evaluator.py` carry the `full` marker and are skipped without it.

### How was this tested?

- `pytest validator/tests -n4 --dist worksteal -p no:cacheprovider` inside `ghcr.io/swarm-subnet/swarm:base`, the command CI runs: **1216 passed, 97 skipped, 0 failed**.
- The command above with `--run-full`, plus `test_run_task.py`: **89 passed**.
- `pre-commit run --all-files`: every hook passed.
- New test `test_run_process_parallel_hands_a_timed_out_seed_back_in_seed_flow`: on the previous code it fails with `assert {0: 2} == {0: 1}`, the seed flown twice. With the change it passes, the result carries `INFRA`, and no retry is logged. `test_run_process_parallel_retries_wall_timeout_once` still passes, so batch flow keeps its retry.
- Two new tests on the claim reply: only repeat hand-outs are reported, and a reply without the field reports nothing.

Validator operators: a frozen seed frees its worker after one timeout instead of two, and a seed that has been around before is named in the log with how many hand-outs it has left. Miners: no scoring change. A timeout was never uploaded as a score and still is not.

### Risk & Rollback

A seed that times out once for a passing reason, such as a loaded host, used to be absorbed locally and now costs one of its six hand-outs on the backend. Before this change the retry was capped at 10 per run anyway. Rollback is a revert of this commit; it has no dependency on the backend PR being deployed.

### Documentation

- [x] Not applicable (explain why): internal evaluator behaviour, not described in `docs/` or the README.


swarm/validator/docker/docker_evaluator_parts/parallel.py
```diff
@@ -880,7 +880,9 @@ def _more_work_expected(done_count: int) -> bool:
                 prior_retries = int(batch_retry_counts.get(int(request.batch_index), 0))
                 if (
                     (
-                        bench_engine._is_timeout_retry_status(final_status)
+                        # In seed flow the pool is the retry: a timed-out seed goes back at
+                        # once instead of holding this worker for a second full timeout.
+                        (bench_engine._is_timeout_retry_status(final_status) and not feeder_active)
```
robojules, line 885: Where is this variable being set?

robojules (changes requested):
PTAL

swarm/validator/docker/docker_evaluator_parts/parallel.py
S3333F: Line 256, at the top of the same function: `feeder_active = seed_feeder is not None`.

It is true only in seed flow, where the backend hands seeds out one at a time and a returned seed is picked up by another validator within a poll or two. Batch flow has no pool to fall back on, so its in-place retry is unchanged.

robojules: Perfect. thanks for the clarification

robojules (approved):
LGTM 🚀

PR #162, opened 2026-09-20
Stop the seeds in flight when a task is cancelled

S3333F:
## Description

When the backend cancelled a task, the validator dropped the seeds that had not started but let every seed already flying run to its natural end. Each seed flies in its own worker process, and the dispatcher had no way to reach a worker, so a cancel took as long as the slowest seed in flight: between **3 and 20 minutes** in the owner validator's log, with the 600 s batch timeout as the ceiling. The dispatcher now shares one stop signal with its workers, so a cancel ends the seeds in flight within a simulation step.

### Related Tasks

- Task ID: OSvPakmN

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Dispatcher | Shares a stop signal with every worker and raises it on a cancel | Operators: a cancel takes seconds |
| Seed batch | Watches the signal while the container starts and while the seed flies | Nobody |
| Retries and scoring | A seed ended by a stop is never retried and keeps no result | Nobody |
| Early phases | The cancel is checked after the model download, the image build and the calibration | Nobody |

- **Flight**: the step loop already checked a stop event on every step, but only the batch timeout ever set it. The batch now sets it when the dispatcher's signal is raised, waits up to **5 s** for the loop to leave the step it is in, and the existing cleanup kills the container.
- **Container start**: a request picked up after the stop launches nothing, not even a spare container. A stop that lands while the agent is still loading ends the wait and removes the container.
- **Fallback**: a worker that has not answered **30 s** after the stop is killed. After any stop the evaluation containers of that UID are swept, which also covers a spare container owned by a killed worker.
- **No retry**: `seed_cancelled` is a timeout-retry status, so a stopped seed used to be eligible for a second flight and consumed the shared retry budget. Retries are now skipped once a stop is in force.
- **No score**: a seed stopped in flight is left without a result, exactly like a seed that never started. It adds no row to the upload, no zero to the running average and no progress tick, and it drops out of the in-flight report so the pool can lease it again.
- **Kept on purpose**: a seed that lands its real result after the stop still counts, finished scores still upload, and the wrap-up submission after a finalize broadcast is unchanged.
- The model download is not interrupted mid-way. It writes a partial file that only the next download removes, so cutting it would leave private bytes on disk. The cancel is checked right after it instead.

### Run it yourself

```bash
```

The cancel paths alone:

```bash
pytest validator/tests/test_container_prewarm.py validator/tests/test_docker_evaluator.py validator/tests/test_run_task.py validator/tests/test_streaming_phase.py -k "stop or cancel or honours"
```

### How was this tested?

- Validator suite in the test image, the CI command: **1225 passed, 96 skipped**.
- Miner suite: **47 passed**.
- `pre-commit run`: all hooks pass.
- **11 new tests**, one per long phase plus retry, scoring, lease hand-back and the kill fallback. **10 of them fail on the base branch**; the eleventh guards the upload behaviour that was already correct.
- A real run on a 12-core host: the baseline model, 24 seeds, 4 workers, the stop raised with 4 seeds in flight, same input on both builds.

| | Base branch | This PR |
|---|---|---|
| Stop requested until the run returned | **263.8 s** | **1.9 s** |
| Containers left behind | 0 | 0 |
| Zero scores recorded for stopped seeds | 0 | 0 |

One case in the owner validator's log took 20 minutes, longer than a single batch timeout. The log is silent for that window and the case was not reproduced, so its cause is not established. The kill fallback bounds any stop at about 30 s regardless.

Validator operators see a cancelled task end in seconds, with the log line `stopped N.Ns after the request`. Miners see no difference: a stopped seed is flown again from the pool and no partial flight is scored.

### Risk & Rollback

If the stop signal were raised wrongly, seeds in flight would be abandoned and flown again after their lease expires: wasted compute, never a wrong score, since a stopped seed uploads nothing. Revert the commit; the backend is untouched.

### Documentation

- [x] Not applicable (explain why): internal validator behaviour, no operator setting or interface changed.


robojules (approved):
LGTM

PR #163, opened 2026-09-20
Keep Docker cleanup off the event loop, bounded and scoped to its owner

S3333F:
## Description

The validator ran its Docker cleanup directly on the event loop after every task, so nothing else could run until Docker answered. Several of the calls had no timeout, including both prunes, so a wedged daemon froze the validator for good with nothing in the log and no restart, because the process stayed alive. The container filters matched by name and the prunes acted on the whole host, so two validators sharing one daemon removed each other's running containers.

Cleanup now runs in a worker thread, every Docker call is bounded, and it only touches containers and images that carry this validator's own label.

### Related Tasks

- Task ID: PTgcThtZ

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Forward loop | Cleanup runs through `asyncio.to_thread`; the loop keeps serving heartbeats, SSE events and uploads | Operators: no frozen validator |
| Timeouts | **30 s** on list and remove calls, **120 s** on the image prune; the first timeout ends the pass with one warning naming the command | Operators: a wedged daemon is visible in the log |
| Ownership | Every evaluation, install and verify container gets a `swarm.instance=<hotkey>` label, and cleanup removes only its own | Anyone sharing a daemon |
| Prunes | `volume prune` and `container prune` removed; `image prune` filtered to `label=swarm.code_hash` | Anyone sharing a host |
| Frequency | After a task: the container sweep and the model image reap. Startup and epoch change: the prunes too | Nobody |
| Telemetry | `cleanup_count`, `last_cleanup_duration_sec` and `last_cleanup_reason` are now written; they existed and were never filled | Operators: the dashboard line shows real values |

- The forward loop still awaits the cleanup before it polls for the next task, so a sweep never overlaps a running task's containers.
- The evaluator never creates a Docker volume, so `volume prune` could only ever delete something that was not ours.
- Model images committed from a labelled install container inherit the label, which is what scopes the image reap. An image another validator still runs from is refused by `docker rmi` as before.
- Startup and the base image rebuild also take unlabelled containers, the ones a release from before the label left behind. After a task only labelled containers are touched.
- The owner name lives in an environment variable so forked and spawned evaluation workers label their containers the same way. Tools with no hotkey (benchmark CLI, video script) use `local`.
- The four unscoped, untimed `docker kill` / `docker rm` shell lines at validator startup are gone; the scoped cleanup that runs right after covers them.
- The label helpers sit in `swarm/utils/docker_instance.py` because the model verifier needs them, and importing them from the validator package would close an import cycle.

### Run it yourself

```bash
docker compose -f .docker/docker-compose.yml up --build swarm_validator
```

The new tests alone, inside the same image:

```bash
pytest validator/tests/test_docker_cleanup.py -n 0 -p no:cacheprovider
```

### How was this tested?

- `pytest -v validator/tests -n4 --dist worksteal -p no:cacheprovider` in the test image: **1224 passed, 96 skipped**.
- `pytest -v miner/tests -n 0 -p no:cacheprovider`: **47 passed**.
- `pre-commit run --all-files`: clean.
- Ten new tests: only the owner's containers are removed, unlabelled ones are taken only when asked, every Docker call in a full pass carries a timeout, no prune after a task, the prune is limited to the Swarm label, model images are listed by owner, the pass ends at the first timeout, launched containers carry the label, and the event loop keeps ticking while a slow cleanup runs.
- Measured against a real daemon (Docker 29.6.2, 12 cores), the old command sequence before and the new function after:

| Case | Before | After |
|---|---|---|
| Nothing to clean | 0.25 to 0.29 s, loop blocked | 0.18 to 0.24 s, loop free |
| 6 leftover containers | 4.42 s, loop blocked | 2.38 s, loop free |
| Another validator's 6 live containers on the same daemon | **all 6 removed** | **all 6 untouched** |
| Daemon paused with SIGSTOP | list call never returns | pass ends at the timeout, one warning, longest event loop gap **53 ms** |

The paused daemon run used a 5 s timeout to keep the probe short; the shipped value is 30 s.

The effect on missed heartbeats on a live validator could not be measured before the change is deployed. The new telemetry fields are what will show it.

Validator operators: no action needed. The first start after the update removes leftover unlabelled containers once. A warning starting with `Docker cleanup abandoned` means the Docker daemon did not answer in time.

### Risk & Rollback

If the label were missing from a container, that container would not be swept after a task and would stay until the next validator start, which takes unlabelled ones. Containers also run with `--rm`, so a normal exit still removes them. Revert the commit to undo; no state or config is migrated.

### Documentation

- [x] Not applicable: no operator step, config or CLI changes.


robojules (approved):
LGTM 🚀

PR #164, opened 2026-09-20
Stop the validator reading a backend failure as a backend answer

S3333F:
## Description

The validator's two signed senders never raised. A timeout, a 5xx or a rate limit came back as an ordinary dict carrying an `error` key, and every caller had to remember to look for it. The seed claim did not: a failed claim has no `granted` and no `pending`, which is exactly what an empty pool looks like, so **one backend hiccup ended the task with seeds unflown**. The task result had the same gap: it was sent once and its reply was never read.

The senders now raise, so a failure is a different kind of thing from an answer and cannot be parsed as one. Each call then gets an explicit rule for an outage and for a refusal.

### Related Tasks

- Task ID: SJfbyx0b

### Type of Change

- [x] Bug fix

### What does this PR change?

| Call | Before | Now |
|---|---|---|
| Signed senders | every failure returned as a dict with an `error` key | raise `BackendTransportError` on an outage (timeout, dropped connection, **5xx**, **429**, unreadable 2xx body) and `BackendRejectedError` on a refusal (**4xx**, with the status and the backend's reason) |
| Claim seeds | any failure read as a drained pool, task ended | an outage keeps the task and asks again on the engine's existing **15 s** wait; gives up after **10 min** of continuous outage; a refusal ends the feeding at once |
| Task result | one send, reply never checked | an outage is retried **8 times** with backoff from 2 s capped at 60 s, about 3 min in total; a refusal or a `recorded: false` reply is logged with its reason |
| Seed scores, refused | resent about **12 times** | sent **once**, dropped, and its seeds leave the in-flight list so the backend can hand them on |
| Seed scores, outage | one last try after the final seed, then lost | parked batches are retried with growing waits for up to **3 min** after the final seed |
| Startup announce | a failed heartbeat read as settled | retried through an outage for the same **4 min** window as the duplicate-session check, then startup carries on |

- A claim reply is only allowed to declare the pool empty when it states `granted` and `pending`. A 200 without them is treated as "ask again".
- The shutdown hand-back needed no edit: it already logs success only when the call returns, and the call now raises when the heartbeat did not land.
- Sync, next-task, the event stream, epoch publish and the progress heartbeat already handled failure correctly and are unchanged. Sync and publish catch the new errors through their existing `except` paths and fall back exactly as before.
- The five timing values are named constants next to the other backend timings.
- Six tests that asserted the old contract itself ("a 502 comes back as a dict tagged `transport_failure`") are removed. `test_backend_failure_handling.py` covers the same cases under the new contract.

### Run it yourself

```bash
```

Or only the files this touches, in the test image:

```bash
docker run --rm -v "$PWD":/app -w /app -e PYTHONPATH=/app -e HOME=/tmp ghcr.io/swarm-subnet/swarm:base \
  pytest -q -n4 -p no:cacheprovider \
  validator/tests/test_backend_failure_handling.py validator/tests/test_run_task.py \
  validator/tests/test_streaming_phase.py validator/tests/test_seed_score_upload.py \
  validator/tests/test_backend_api.py
```

### How was this tested?

- `pytest -v validator/tests -n4 --dist worksteal -p no:cacheprovider` in the test image: **1255 passed, 96 skipped**.
- `pre-commit run --all-files`: every hook passed, including the docstring check.
- **49 new test cases**: each failure kind (500, 502, 429, read timeout, refused connection, HTML under a 200, a JSON list under a 200) against both senders, the refusal statuses, and the claim, result, score upload and startup rules above, including the give-up points.
- The real client over real HTTP against a local server that fails and then recovers, the same script on `main` and on this branch:

| Scenario | `main` | This branch |
|---|---|---|
| Claim answered with 500 | `drained=True`, task ends | raises, task kept |
| Claim answered with 429 | `drained=True`, task ends | raises, task kept |
| Claim answered with a real grant | seeds 4 and 9 granted | seeds 4 and 9 granted |
| Result answered with 503, 503, 200 | **1 send**, error dict returned, result lost | **3 sends**, `recorded: true` |

Validator operators: a short backend outage no longer makes a validator drop a model mid-evaluation or lose a finished result. Logs now say whether the backend was unreachable or refused, with the reason. No config change and nothing to do. Scores, weights and the wire protocol are untouched, and the backend needs no change.

### Risk & Rollback

If a rule is wrong the cost is time, not scores: a validator could wait on a dead backend for up to 10 min inside one task before moving on, where it used to leave at once. Revert the commit to undo it; nothing is stored and nothing on the backend depends on it.

### Documentation

- [x] Not applicable (explain why): internal error handling of the validator's backend client, no CLI, config or protocol surface changed.


robojules (approved):
LGTM 🚀

PR #165, opened 2026-09-20
Keep the per-seed timing record and time every phase of a seed

S3333F:
## Description

Every seed builds a timing record in the flight loop: its wall time, the calibration overhead and the CPU factor. The record crossed the worker process and reached the dispatcher intact, then died at the last hop: the heartbeat's seed callback took no argument, the call raised a `TypeError`, and the emitter caught it and called again with nothing. That happened on every seed, silently, so no timeout or retry value could be chosen from evidence. The callback now accepts the record, every phase of a seed has its own clock, and the dispatcher writes each attempt down where it can be read per seed and in aggregate.

### Related Tasks

- Task ID: c4YOhMIV

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Heartbeat | The seed callback accepts the record, so the `TypeError` fallback never fires | Nobody |
| Flight loop | Clocks for env build, reset, calibration, flight, miner time, simulator time and cleanup | Nobody |
| Seed batch | Adds container start, connect time and the last phase reached to every record | Nobody |
| Dispatcher | Stamps queue wait, total and attempt number, and logs every attempt | Operators: one summary line per run |
| Telemetry | A capped per-seed log, a rolling summary in the snapshot, a section in `swarm monitor` | Operators |

- **The bug**: `HeartbeatManager.on_seed_complete` takes an optional record. Progress counting is unchanged, it still only counts seeds.
- **Existing fields**: nothing is renamed or dropped. The record that reaches the log carries every field the flight loop already built.
- **Flight**: `fly_sec` is the step loop. Inside it `act_sec` is the time spent waiting on the miner, `act_max_sec` its slowest step, and `sim_sec` the time inside `env.step`. The three together say whether a slow seed is the miner or the simulator.
- **Cleanup**: the record is held until the environment is closed, so `cleanup_sec` is inside it. The progress event therefore arrives after teardown instead of before it, a fraction of a second later.
- **Container**: setup, launch, lockdown and serve were already measured and printed as one worker log line. They now ride on the record, with `connect_sec` and whether the container was prewarmed. `serve_sec` is also set when the agent never answers, so a failed start reports how long it waited.
- **Attempts**: a seed that times out and is retried used to leave no trace of its first attempt, because the retry path skips the callback. The dispatcher now logs that attempt with `retried: true` before sending it round again, and logs a batch whose worker died. The progress callback still fires once per final seed.
- **Where it lands**: `state/validator_seed_timing.jsonl`, one line per attempt, rolled over to a single `.1` file at **20 MB**. It is a separate file because the existing event log has no rotation and the monitor re-reads it on every refresh. The snapshot keeps count, median, p90 and max per phase over the last **2000** seeds, plus the total split by status.
- **Log line**: each run ends with one `seed timing n=...` line, so the numbers are readable from a validator's console log without access to the box.
- Nothing is sent to the backend.

### Run it yourself

```bash
```

The timing tests alone:

```bash
pytest validator/tests/test_seed_timing.py
```

### How was this tested?

- Validator suite in the test image, the CI command: **1224 passed, 96 skipped**.
- Miner suite: **47 passed**.
- The opt-in engine and telemetry tests with `SWARM_RUN_FULL=1`: **155 passed**.
- `pre-commit run --all-files`: all hooks pass.
- **10 new tests**: a flown seed, a seed whose env breaks twice, the container fields, a normal seed through the dispatcher, a timed-out seed and its retry, a dead batch, the aggregate, the rollover, the log line and the monitor section.
- On the base branch `hb.on_seed_complete({...})` raises `takes 1 positional argument but 2 were given`. On this branch a real run delivered **9 of 9** records to the callback and none as a bare call.
- A real run on a 12-core host, real containers, the `cf_swarm_sar` family, 3 workers, one normal model and one deliberately slow model.

| Phase, normal model, 6 seeds | Median | Max |
|---|---|---|
| **Total per seed** | **2.8 min** | 4.2 min |
| Simulator (`sim_sec`) | **2.1 min** | 3.3 min |
| Miner (`act_sec`) | 24.5 s | 27.0 s |
| Env build | 4.0 s | 9.8 s |
| Container start | 1.5 s | 1.9 s |
| Connect | 53 ms | 4.0 s |
| Cleanup | 260 ms | 599 ms |
| Queue wait | 6 ms | 210 ms |

| Slow model, 3 seeds | Status | Total | Miner time |
|---|---|---|---|
| Seed 1 | `seed_timeout_strikes` at step 39 | **66 s** | 32 s |
| Seed 2 | `seed_done` | **905 s** | 801 s |
| Seed 3 | `seed_done` | **1042 s** | 805 s |

<details>
<summary>The summary line of the slow run</summary>

```
[benchmark] UID 191 seed timing n=3 (seed_done 2, seed_timeout_strikes 1) median/p90/max sec | total 904.75/1042.01/1042.01 | queue_wait 0.01/0.02/0.02 | container_start 2.22/2.34/2.34 | connect 14.12/14.15/14.15 | seed_wall 887.26/1024.55/1024.55 | env_build 9.22/10.68/10.68 | reset 0.00/0.01/0.01 | calibration 0.04/0.05/0.05 | fly 886.33/1015.30/1015.30 | act 800.57/804.98/804.98 | act_max 0.92/1.26/1.26 | sim 80.42/204.62/204.62 | cleanup 0.43/0.44/0.44
```

</details>

The retried-attempt path is covered by a test only. No seed was retried in the real runs.

Validator operators get a Seed Timing section in `swarm monitor`, one summary line per run in the log, and a new file under `state/` capped at 40 MB across its two parts. Miners see no difference: no score, timeout or retry value changes.

### Risk & Rollback

The clocks add two `perf_counter` reads per simulation step and the tracker call is wrapped so a failure in it cannot reach the evaluation. If the held record were wrong, a progress tick would arrive late, not go missing: the dispatcher already rebuilds a record from the result when none arrived. Revert the commit to undo it; the new file can be deleted at any time.

### Documentation

- [x] Not applicable (explain why): the record is internal telemetry and the monitor section labels itself.


robojules (approved):
LGTM 🚀

PR #166, opened 2026-09-20
Give the seed clock room for the thinking the per-step budget allows

S3333F:
## Description

A miner may spend **600 ms** on every step, and a seed is 3,000 steps, so the rules allow up to 1,800 s of thinking per seed. The seed clock was **840 s** on autopilot, thinking and simulation together. A model that thinks 250 ms per step breaks no rule and still cannot finish a full-length flight on a village or mountain map: it hits the clock, the validator books it as its own infrastructure fault, and the seed goes back to the pool. In production over 2.3 days that was **86 seeds returned with no result, a median 1,701 s each, the same seeds across three models and four validators**, all village and mountain.

The clock now holds the thinking the per-step budget already allows. On top, every uploaded score carries the timings the validator measured, and the engine's shape warning no longer floods the log.

Stacked on #165, which adds the per-seed timing this reads. It targets that branch so the diff shows only this change; once #165 lands it retargets to `v5.1.6.1`.

### Related Tasks

- Task ID: srRmWgbZ
- Related tasks/PRs (other repos): #165 (the timing record this PR reads; merge first), #161 (hands a timed-out seed back instead of flying it twice; independent)

### Type of Change

- [x] Bug fix

### What does this PR change?

|---|---|---|
| Seed clock | batch timeout = family clock **+ steps x 600 ms x host speed factor**, capped at the family ceiling | Operators: slow legal models finish instead of looping |
| Score upload | each row carries `runtime_details`: speed factor, thinking, slowest step, simulation, build, wall, total, attempt | Nobody; the backend column already exists and was empty |
| Map tagging | the engine's `getCollisionShapeData failed` warning is muted on stdout as well as stderr, with the C buffers flushed | Operators: **10.26 M** junk lines gone from the owner log |

- **Clock.** `_legal_thinking_sec` sums `ceil(horizon / sim_dt)` over the batch and multiplies by `MINER_COMPUTE_BUDGET_SEC` and `max(1.0, speed_factor)`. Autopilot on a 1.0x host goes **840 s -> 2,640 s**, on a 1.3x host **3,180 s**; the family ceiling (3,600 s) still binds above that. Statuses, retries and counting are untouched. The stall cut on the card 194 branch lives in the wait loop below this expression and merges cleanly (trial merge, full suite 1232 passed).
- **Upload.** The scheduler attaches the timing record to the result before the result callback fires; the streaming phase copies eight fields into the row only when a record exists, so rows from paths without timing are byte-identical to before.
- **Tagging.** The engine prints the warning through C stdio on stdout, block-buffered under pm2, so the old stderr-only mute never caught it. Warehouse maps have **4,442** shelf pieces without a collision shape, one warning each.

### Run it yourself

```bash
```

### How was this tested?

- Validator suite in the CI image, `pytest validator/tests -n4 --dist worksteal`: **1228 passed, 0 failed, 96 skipped** (opt-in heavy tests).
- Four new tests, all failing on the base commit and passing here: the allowance arithmetic, the timing record on the result, the upload row's `runtime_details`, and a warehouse tagging pass that leaves both output descriptors empty.
- `pre-commit run --all-files` clean.
- Real evaluator on rendervps, 6 workers, host speed factor 1.30, a test agent that thinks 256 ms per step (zero strikes, slowest act 0.38 s), on the production seeds that timed out:

| Seed | Before (840 s clock) | After |
|---|---|---|
| Mountain, index 556 | `Batch timeout after 840.1s` at step 2,648 of 3,000, booked as infra, retried | finished in **964 s**, scored once, no retry |
| Village, index 493 | finished in 821 s, 19 s under the line | finished in 823 s, scored once |

- Warehouse build with every body classified: **377,570 bytes** of engine warnings on stdout before, **0** after.

Validator operators: a slow but rule-abiding model now completes its seed on the first attempt instead of holding a worker for two clocks and coming back unscored. Scores do not change; a seed that finishes is scored exactly as it was. No config change.

### Risk & Rollback

A seed whose simulation freezes now holds a worker up to the family ceiling instead of the old family clock; the hard cap on `act()` still catches a hung model early, and the stall cut on the card 194 branch closes the remaining gap. Revert the commit; nothing else depends on it.

### Documentation

Nothing to update: no operator-facing option or miner-facing rule changed.


robojules (approved):
LGTM 🚀

PR #167, opened 2026-09-20
Cut a seed that stops reporting progress instead of waiting for the batch limit

S3333F:
## Description

A seed in flight is watched by one clock: the family's batch limit, **840 s** on autopilot. It was never checked against how long a seed takes. Over the last three days, across all five validators, it cut **77 healthy seeds**, every one of them a village or mountain seed of a legal model that was still flying; each cut held a worker for **1,703 s** (two attempts) before the seed went to another validator, which flew it fine in **535 s** median. That is **40.6 worker-hours**, 8% of all autopilot flying time. Those 77 were slow, not frozen, and the clock that cut them is the one #166 raises: with the per-step thinking allowance the autopilot limit becomes **2,640 s** on a 1.0x host, which clears the slowest measured mountain seed (1,028 s) with room.

What no clock caught in those three days is the other case: a seed that freezes. A frozen seed and a slow seed look the same to a wall-clock limit, and with #166 a frozen one would now hold a worker for up to the raised limit. They look different to the progress stamps the flight loop already writes at every phase and every step. The batch now watches those: a seed that has not stamped progress for **222 s** is cut at once, and the batch limit stays as the ceiling behind it. Sits on top of #166.

### Related Tasks

- Task ID: eZo6R5hq
- Related tasks/PRs: https://github.com/swarm-subnet/swarm/pull/166 (raises the batch clock this PR sits behind; merged first)

### Type of Change

- [x] Bug fix

### What does this PR change?

- **Stall clock**: `SEED_STALL_TIMEOUT_SEC = 222`. In the batch wait loop, when `progress_state["ts"]` is older than that, measured from the start of the RPC phase, the seed is cut through the existing timeout path. Status, retry and hand-back behaviour are the ones a batch timeout has today, so the dispatcher, the seed feeder and the backend see nothing new.
- **Log line**: a stall is reported as `Seed stalled for UID n: no progress for Xs (stall_limit=222.0s, elapsed=…, phase=…, step=…)`, so it can be told apart from a batch limit in a validator's console log. The batch limit line is unchanged.
- **Nothing else moves**: the family batch limits, the per-step budgets, the strike counts, the connect and start clocks, the parent heartbeat and the scoring are untouched.

Why 222 s: the longest a healthy seed goes without a progress stamp is a map build or a cold connect. Measured on a 12-core host with 6 workers and the current champions, the worst was **14.1 s** (cold connect), the worst map build **12.8 s** (warehouse). The bounded waits inside a seed are smaller still: reset 5 s, calibration 5 s, a first step 9 s at the slowest eligible host. A test pins the constant above every one of those and below the smallest batch limit.

### Run it yourself

```bash
```

The stall clock alone, in the test image:

```bash
docker run --rm -v "$PWD":/app -w /app -e PYTHONPATH=/app -e HOME=/tmp ghcr.io/swarm-subnet/swarm:base \
  pytest validator/tests/test_seed_stall_timeout.py -n 0 -p no:cacheprovider
```

### How was this tested?

- Validator suite in the test image, the CI command: **1218 passed, 96 skipped**.
- Miner suite: **47 passed**.
- `pre-commit run --all-files`: every hook passes, including the docstring check.
- Four new tests through the real worker body with the in-memory docker daemon: a flight that never stamps progress is cut at the stall limit (**0.6 s** elapsed with the limit set to 0.5 s) with status `batch_timeout_partial` and score 0; a flight that stamps every 0.1 s for three stall limits is never cut and keeps its score; a flight that keeps stamping is still cut by the batch limit; and the constant sits above every bounded wait a seed can make.
- Measured on a 12-core host, 6 workers, the current champion of each family flown through `evaluate_seeds_parallel` with the batch limit lifted so nothing was cut. Longest quiet phase per family, the number the stall limit has to clear:

| Family | Seeds | Map build max | Cold connect max | Seed total median | Seed total max |
|---|---|---|---|---|---|
| Autopilot | 120 | **12.8 s** | **14.1 s** | 134 s | **1,028 s** (mountain) |
| Office interceptor | 24 | 3.7 s | 8.1 s | 36 s | 103 s |
| Search and rescue | 24 | 8.9 s | 4.1 s | 133 s | 372 s |

<details>
<summary>Autopilot per map, where the time goes</summary>

| Map | Total median | Total p90 | Total max | Miner time median | Simulator median | Map build max |
|---|---|---|---|---|---|---|
| mountain | 592 s | 936 s | 1,028 s | 387 s | 204 s | 11.4 s |
| village | 343 s | 394 s | 411 s | 305 s | 28 s | 5.3 s |
| warehouse | 142 s | 472 s | 572 s | 84 s | 55 s | 12.8 s |
| open | 107 s | 148 s | 170 s | 76 s | 21 s | 1.7 s |
| city | 98 s | 112 s | 150 s | 75 s | 9 s | 3.2 s |
| forest | 75 s | 112 s | 136 s | 20 s | 43 s | 4.1 s |

Four of the twenty mountain seeds ran past 840 s. Nothing in them was quiet: the miner and the simulator were stamping every step.

</details>

- Production, `seedattempt` rows 2026-09-17 to 09-20, five validators, all families: **27,516** healthy seeds; **77** returned without a result, all autopilot village or mountain, median hold **1,703 s**; re-flown elsewhere they settled in **535 s** median. Healthy autopilot p99 per map: mountain 780 s, village 889 s, every other map under 430 s.

Validator operators: a frozen seed frees its worker after **222 s** instead of the family batch limit (840 s to 2,730 s), and the log names it as a stall. Miners: no scoring change; a stalled seed is handled exactly as a timed-out seed is today.

### Risk & Rollback

If a real phase ever stayed quiet longer than 222 s on some host, that seed would be cut as a stall instead of running on. No such phase exists in the measurements or in the production history, and the cut lands on the same retry path as a batch timeout, so the seed is not lost. Revert the commit to undo it.

### Documentation

- [x] Not applicable (explain why): the constant is documented where it is defined and the log line names itself.


robojules (approved):
LGTM 🚀

PR #170, opened 2026-09-21
Add the solar park map builder with seed rules and a movers runtime

S3333F:
## Description

The Manolas solar park exists as a frozen Blender scene and has no way into the simulator: every other map is generated in code, and a real site with fixed rows, a fence and a road cannot be. This adds a builder that places a map from a manifest of exported pieces, lets the seed vary only what may vary on a real site, and drives the three movers a step at a time. The assets themselves ship in `swarm-worlds` under `maps/custom/solar`; without them the builder raises on the missing manifest and the tests skip.

### Related Tasks

- Task ID: 4DVIDEAR

### Type of Change

### What does this PR change?

|---|---|---|
| Map builder | `swarm/core/maps/solar/builder.py` places every item and placement the manifest lists, with the visual flags and the collision kind each item declares | Nobody yet: no family calls it |
| Seed rules | Vegetation density per tier, movers' timing and paths, all as numbers in one `CONFIG` | Nobody yet |
| Movers runtime | `SolarMovers` moves the pickup, the buzzard and the goat herd one 50 Hz step at a time | Nobody yet |
| Tests | `validator/tests/test_solar_movers.py`, 11 tests against the real assets, skipped when they are absent | CI |

- **What the seed decides.** The layout never moves. The seed draws the share of trees and grass that stands per tier (merged pieces carry a `band_low` rank so a whole band stands or falls together), whether the truck is parked or when it leaves (**0 to 40 s**), where the bird starts on its loop, and the herd's start offset, headings, turn rates and paces from a herd seed. Everything is in `CONFIG` and nothing else.
- **Pickup.** Body, glass and four wheels follow a 50 Hz table; wheels sit at their local pose in the body frame and spin by the table's angle column; the truck holds the last row at the dead end.
- **Bird.** Eighteen parts follow the carrier path composed with a per-part wingbeat table, index wrapped on the loop.
- **Goats.** Each walks the arc its constant turn rate draws, its foot dropped on the terrain by a downward ray every step (**4 us** a ray), its mesh rewritten with `resetMeshData` from the stride at a phase that never stops moving, the two neighbouring frames blended. Stride length, frame count and rate are read from the pose table, never assumed. An engine that refuses `resetMeshData` on a rigid mesh gets the herd standing still, without raising.
- **Multi-material pieces.** Items flagged `materials_from_mtl` load with `VISUAL_SHAPE_MATERIALS_FROM_MTL`, which is how a merged cell of many trees keeps ten textures in one body.
- The `pickup_body` carries the only collision box among the movers and classifies as `SUPPORT_ROOFTOP` under challenge types 1 and 4, so a family taking this map must tag `movers.body_uids` before `build_and_tag_map`. A test pins it.

### Run it yourself

```bash
```

With the assets in place, the map alone:

```bash
SOLAR_ASSET_DIR=/path/to/maps/custom/solar pytest validator/tests/test_solar_movers.py -q
```

### How was this tested?

- `pytest validator/tests/test_solar_movers.py -q` against the exported assets: **11 passed in 13.6 s**. Covered: every mover body held; stride facts read off the pose table; blend endpoints and wrap; the truck waits its delay, walks the table and holds the last row; wheels keep their place and turn by exactly the spin column; the bird wraps; goats walk with feet within **5 cm** of an independently rayed ground over 100 steps (measured **0.0 cm**), and the test demands the terrain actually moves under them; two seeds walk the herd apart, one seed repeats itself.
- `pre-commit run --all-files`: **all 14 hooks pass**.
- The whole map through this builder on the Swarm engine fork, seed 7, 2 render threads: **3,117 bodies, 3.26 M placed triangles, build 27.8 s, 256 px sensor frame 81 ms, peak RSS 2,416 MB**.
- `advance()` with the 29 mover bodies and no herd mesh: **0.15 ms** median. Rewriting five goat meshes costs about **150 ms** a step, four fifths of it converting 52,392 vertices into the nested list the binding accepts; that cost is the engine binding's and is the open point before a family uses the herd at every step.

### Tools & Dependencies

- Tools updated: none in this repository. The assets and the export scripts that produce the manifest live outside it.

### Risk & Rollback

No family or challenge type references the builder, so nothing in production changes. Revert the three commits to remove it.

### Documentation

- [x] Not applicable: the module docstring carries how the manifest and the seed rules work, and no user-facing document describes maps.


robojules (approved):
LGTM

PR #171, opened 2026-09-21
Bring the export checker's material and texture rules up to date with the engine

S3333F:
## Description

The export checker still enforced two rules the engine no longer has. An OBJ with more than one material was an error, although a piece loaded with `VISUAL_SHAPE_MATERIALS_FROM_MTL` keeps every material apart; and a PNG with an alpha channel was an error, although the decoder keeps the alpha plane and `ER_ALPHA_CUTOUT` reads it, which is how every leaf and fence cut-out works. On the solar park export those two rules produced **21 of its 22 errors**. A third rule called a flat double-sided card an inside-out shell whenever its signed volume rounded to a negative zero.

### Related Tasks

- Task ID: 4DVIDEAR

### Type of Change

- [x] Bug fix

### What does this PR change?

- **Materials**: more than one `usemtl` in a file is a warning that names the flag which keeps them apart, with the split as the alternative fix.
- **Textures**: the alpha branch is gone, including the `tRNS` sniff that caught palette transparency for the same stale reason. 16-bit PNG and progressive JPEG still fail, and a 16-bit RGBA PNG is now covered by a test so removing the alpha rule cannot open a hole.
- **Orientation**: a closed shell is inside out only when its signed volume is negative beyond **1e-6** of the volume its faces sweep, so a two-triangle card wound both ways, whose true volume is zero, no longer trips on floating-point noise.
- The module docstring and the scripts README say what the rules are now.

### Run it yourself

```bash
pytest validator/tests/test_check_blender_export.py -q
python3 validator/scripts/check_blender_export.py <folder of an OBJ export>
```

### How was this tested?

- Each of the three changes was written as a failing test first: multi-material asserted as an error on `main`, RGBA PNG asserted as an error on `main`, and the flat card asserted inside out on `main` with corners chosen so the volume rounds to **-9e-16**. After the change: `pytest validator/tests/test_check_blender_export.py -q`, **20 passed**.
- `pre-commit run --all-files`: **all 14 hooks pass**.
- On the solar park export (park, terrain, plants, grass: 4,300 OBJ files): **22 errors on `main`, 0 after**. The warnings are unchanged, open edges on cut-out cards and ground tiles by design.

### Risk & Rollback

A real multi-material piece loaded without the flag now passes with a warning instead of an error, which is the trade for maps that use the flag. Revert the two commits to return to the old rules.

### Documentation

- [x] `validator/scripts/README.md` updated


robojules (approved):
LGTM 🚀

